PR #1, opened 2019-02-15
Add WordPress

joehoyle:
Installing this module should install WordPress. We want WordPress to be installed to the project root, which for now `johnpbloch/wordpress` does without a problem.

Do move the index.php and wp-config.php into their location, we have to make this a composer-plugin, so we get install and update callbacks. The wp-config is mostly a shell, were DB constants etc are meant to be provided by modules that are specific to the current environment, like Chassis, Cloud etc.


inc/Composer/Plugin.php
```diff
@@ -0,0 +1,36 @@
...
+	public static function getSubscribedEvents() : array {
+		return [
+			'post-update-cmd' => [ 'on_post_update_cmd' ],
+			'pre-install-cmd' => [ 'on_post_update_cmd' ],
+		];
+	}
+
+	public function on_post_update_cmd() {
+		$source = dirname( dirname( __DIR__ ) );
+		$dest = dirname( dirname( dirname( $source ) ) );
```
rmccue: `dirname( $source, 3  )` would be clearer

inc/Composer/Plugin.php
```diff
@@ -0,0 +1,36 @@
...
+
+use Composer\Composer;
+use Composer\IO\IOInterface;
+use Composer\Plugin\PluginInterface;
+use Composer\EventDispatcher\EventSubscriberInterface;
+
+class Plugin implements PluginInterface, EventSubscriberInterface {
+	public function activate( Composer $composer, IOInterface $io ) {
+	}
+	public static function getSubscribedEvents() : array {
```
rmccue: #needsphpdoc

inc/Composer/Plugin.php
```diff
@@ -0,0 +1,36 @@
...
+		return [
+			'post-update-cmd' => [ 'on_post_update_cmd' ],
+			'pre-install-cmd' => [ 'on_post_update_cmd' ],
+		];
+	}
+
+	public function on_post_update_cmd() {
```
rmccue: Since this is used for multiple events, something more descriptive would be better.

inc/Composer/Plugin.php
```diff
@@ -0,0 +1,36 @@
...
+
+		// Update the .gitignore to include the wp-config.php, WordPress, the index.php
+		// as these files should not be included in VCS.
+		if ( ! file_exists( $dest . '/.gitignore' ) ) {
+			file_put_contents( $dest . '/.gitignore', "wordpress/\nindex.php\nwp-config.php" );
+		}
+
+		@mkdir( $dest . '/content' ); // @codingStandardsIgnoreLine
+		@mkdir( $dest . '/content/plugins' ); // @codingStandardsIgnoreLine
+		@mkdir( $dest . '/content/themes' ); // @codingStandardsIgnoreLine
```
rmccue: `if ( ! file_exists() )` to avoid the asperands and ignores here

wp-config.php
```diff
@@ -0,0 +1,53 @@
...
+if ( ! defined( 'ABSPATH' ) ) {
+	define( 'ABSPATH', __DIR__ . '/wordpress/' );
+}
+
+if ( ! defined( 'WP_CONTENT_DIR' ) ) {
+	define( 'WP_CONTENT_DIR', $_SERVER['DOCUMENT_ROOT'] . '/content' );
+}
+
+if ( ! defined( 'WP_CONTENT_URL' ) ) {
+	define( 'WP_CONTENT_URL', 'http://' . $_SERVER['HTTP_HOST'] . '/content' );
```
rmccue: `https`?

composer.json
```diff
@@ -0,0 +1,24 @@
...
+  "require": {
+    "johnpbloch/wordpress": "5.0.3",
+    "composer-plugin-api": "^1.1"
+  },
+  "extra": {
+    "class": "HM\\Platform\\CMS\\Composer\\Plugin"
+  },
+  "autoload": {
+    "psr-4": {
+      "HM\\Platform\\CMS\\": "inc/"
```
rmccue: Might be better to use a classmap here?

wp-config.php
```diff
@@ -0,0 +1,56 @@
+<?php
+
+/*
+ * Main config file for loading Platform.
+ *
+ * DO NOT EDIT THIS FILE.
+ *
+ * All configuration should be done either in your projec'ts composer.json or `config/`
```
rmccue:
```suggestion
 * All configuration should be done either in your project's composer.json or `config/`
```

PR #2, opened 2019-03-05
Load plugin API early

joehoyle:
Load the WordPress API early, so autoloaded files can make use of add action etc.

### DIFF
--- a/wp-config.php
+++ b/wp-config.php
@@ -9,6 +9,9 @@
  * directory.
  */
 
+// Provide a reference to the app root directory early.
+define( 'ROOT_DIR', __DIR__ );
+
 // Load the plugin API (like add_action etc) early, so everything loaded
 // via the Composer autoloaders can using actions.
 require_once __DIR__ . '/wordpress/wp-includes/plugin.php';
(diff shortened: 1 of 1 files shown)


wp-config.php
```diff
@@ -9,10 +9,16 @@
...
 
+// Load the plugin API (like add_action etc) early, so everything loaded
+// via the Composer autoloaders can using actions.
+require_once __DIR__ . '/wordpress/wp-includes/plugin.php';
+
 // Load the whole autoloader very early, this will also include
 // all `autoload.files` from all modules.
 require_once __DIR__ . '/vendor/autoload.php';
 
+do_action( 'loaded_autoloader' );
```
rmccue: Let's use a namespaced action here, to make it clear that this is in Platform. Something like `hm-platform.loaded_autoloaders`?

wp-config.php
```diff
@@ -9,10 +9,19 @@
  * directory.
  */
 
+// Provide a reference to the app root directory early.
+define( 'ROOT_DIR', __DIR__ );
```
rmccue: Should we namespace this? Maybe `HM\Platform\ROOT_DIR` or something?

joehoyle: I think because it's the constant that is the _project_ root dir, having it in the NS might technically be correct but is less natural. We could call it `PROJECT_ROOT_DIR` to be more descriptive though.

rmccue: I'm thinking we should avoid global namespace where possible, both as a best practice and to avoid conflicts with WordPress.

PR #3, opened 2019-03-06
Remove updates from showing in the WordPress admin

joehoyle:
This removes the update nags and cron checks for themes, plugins and core.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,6 +17,10 @@
     "class": "HM\\Platform\\CMS\\Composer\\Plugin"
   },
   "autoload": {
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -17,6 +17,10 @@
     "class": "HM\\Platform\\CMS\\Composer\\Plugin"
   },
   "autoload": {
-    "classmap": [ "inc/" ]
+    "classmap": [ "inc/" ],
+    "files": [
+      "inc/remove_updates/namespace.php",
```
rmccue: Require this from `load.php` instead IMO

PR #6, opened 2019-03-19
Add branding part of CMS

joehoyle:
This removes from WordPress bits, adds a color scheme for Platform and option to set a logo for login


joehoyle:
Depends on #1

inc/branding/namespace.php
```diff
@@ -0,0 +1,63 @@
+<?php
+
+namespace HM\Platform\CMS\Branding;
+
+function bootstrap() {
```
rmccue: Needs phpDoc

inc/branding/namespace.php
```diff
@@ -0,0 +1,63 @@
...
+}
+
+/**
+ * Remove dashboard widgets that are not useful.
+ */
+function remove_dashboard_widgets() {
+	remove_meta_box( 'dashboard_primary', [ 'dashboard', 'dashboard-network', 'dashboard-user' ], 'side' );
+}
+
+function add_color_scheme() {
```
rmccue: Needs phpDoc

inc/branding/namespace.php
```diff
@@ -0,0 +1,63 @@
...
+			'#152354',
+			'#14568A',
+			'#D54E21',
+			'#2683AE',
+		],
+		[
+			'base' => '#152354',
+			'focus' => '#fff',
+			'current' => '#fff',
+		]
```
rmccue: Could we move these into a constant or something instead?

inc/namespace.php
```diff
@@ -0,0 +1,39 @@
+<?php
+
+namespace HM\Platform\CMS;
+
+use function HM\Platform\get_config;
+
+function bootstrap() {
```
rmccue: Needs phpDoc

inc/namespace.php
```diff
@@ -0,0 +1,39 @@
...
+	require_once __DIR__ . '/remove_updates/namespace.php';
+	require_once __DIR__ . '/branding/namespace.php';
```
rmccue: Can we move these to `load.php` instead? No need to conditionally load them in.

inc/namespace.php
```diff
@@ -0,0 +1,39 @@
...
+
+	if ( $config['login-logo'] ) {
+		add_action( 'login_header', __NAMESPACE__ . '\\add_login_logo' );
+	}
+}
+
+/**
+ * Add the custom login logo to the login page.
+ *
+ * @return void
```
rmccue: I wouldn't bother with this, but if you do, be consistent and use it on all the other functions too.

inc/remove_updates/namespace.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+namespace HM\Platform\CMS\Remove_Updates;
+
+function bootstrap() {
```
rmccue: Needs phpDoc

inc/remove_updates/namespace.php
```diff
@@ -0,0 +1,30 @@
...
+	add_action( 'admin_init', 'HM\\Platform\\CMS\\Remove_Updates\\remove_update_nag' );
+	add_action( 'plugins_loaded', 'HM\\Platform\\CMS\\Remove_Updates\\remove_update_check_cron' );
```
rmccue: `__NAMESPACE__` here instead.

inc/branding/namespace.php
joehoyle: nah

joehoyle: these are just for the svgs, so I don't think it's worth it really

assets/admin-color-scheme.css
```diff
@@ -0,0 +1,504 @@
...
+ */
+ body {
+	background: #f1f1f1;
+  }
+
+  /* Links */
+  a {
+  color: #0073aa;
+  awd: #4667de;
+  awd: #152a4e;
```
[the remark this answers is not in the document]
joehoyle: my bad! ....move along

PR #22, opened 2019-04-11
Add admin bar icon menu

joehoyle:
This "brings back" the wp menu, but actually makes it an HM menu, where we can add jump links etc in the future.


inc/branding/namespace.php
```diff
@@ -115,3 +134,24 @@ function detect_missing_default_theme() {
...
+
+/**
+ * Add the Platform logo menu.
+ *
+ *
+ * @param WP_Admin_Bar $wp_admin_bar
+ */
+function admin_bar_menu( WP_Admin_Bar $wp_admin_bar ) {
+	$logo_menu_args = [
+		'id'    => 'hm-platform-logo',
```
rmccue: Can we make this `hm-platform` instead? Should be clearer when adding subitems.

PR #35, opened 2019-04-17
Add WP User Signups

### LINKED ISSUE #29: Add ability to manage pending registration
New/pending registrations in WordPress currently disappear into a blackhole which is unmanageable by site administrators.

joehoyle:
Fixes #29.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,8 @@
   ],
   "require": {
     "johnpbloch/wordpress": "5.1.1",
-    "composer-plugin-api": "^1.1"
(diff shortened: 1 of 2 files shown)


JJJ:
It is not exactly _not_ maintained, it's just what I needed it to be at the time.

Happy to maintain or add access for anyone that wants it. I have a hunch we will be on the same page for whatever y'all think it needs. 🤗

rmccue:
@JJJ Thanks, suspected as much on both counts!

PR #52, opened 2019-05-10
WordPress 5.2 Modifications

joehoyle:
In https://github.com/humanmade/platform-cms-installer/pull/5 we bumped to WordPress 5.2.

We want to disable a couple of the features, specifically Site Health and the Fatal Handler. These are both features for non-technical users to understand the php errors and environment errors for their site. As we are targetting a much more technical audience of developers who will be expected to deal with these issues, these features are not aligned well with what we want to show in the admin.

The WP Recovery Mode as it happens is not enabled in Multisite, so we don't have to do anything there.

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -31,6 +31,9 @@ function bootstrap() {
 	}
 
 	add_filter( 'pre_site_option_fileupload_maxk', __NAMESPACE__ . '\\override_fileupload_maxk_option' );
+	add_filter( 'wp_fatal_error_handler_enabled', __NAMESPACE__ . '\\disable_wp_fatal_handler' );
+	add_action( 'admin_menu', __NAMESPACE__ . '\\remove_site_healthcheck_admin_menu' );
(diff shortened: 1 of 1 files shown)


inc/namespace.php
```diff
@@ -64,3 +67,42 @@ function load_plugins() {
...
 }
+
+/**
+ * Disable the WP fatal error handler.
+ *
+ * This is intended mostly for non-technical users to receive error information.
+ *
+ * @return boolean
+ */
+function disable_wp_fatal_handler() : bool {
```
rmccue: This is a confusing name. Can we just use `__return_false` instead?

joehoyle: Yeah, turns out that function is not available that early, as the fatal handler is loaded super early

rmccue: OK, let's rename this to `filter_wp_fatal_handler` or something; having a function called `disable_...` return `false` is just a bit weird.

rmccue: Also, let's document this

joehoyle: Ok

PR #54, opened 2019-05-10
Add the WP installer to this package

roborourke:
Will mean we have less updates to do on cms-installer and we can control the WP version with less effort.

Updating to 5.2 at the same time. Question though, should we make this `~5.2.0` to reduce WP upgrading friction?

Related to https://github.com/humanmade/platform-cms-installer/pull/5

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,6 +11,7 @@
   "require": {
     "php": ">=7.1",
     "stuttter/wp-user-signups": "3.1.0",
+    "johnpbloch/wordpress": "5.2.0",
     "humanmade/cms-installer": "0.1.0"
   },
   "extra": {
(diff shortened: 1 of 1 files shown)


rmccue:
> Question though, should we make this `~5.2.0` to reduce WP upgrading friction?

I think so, yes.

joehoyle:
IMO we're better to control the exact version. WordPress has shown to have larger changes in a patch version than probably any other plugin. In situations where there is a larger change, I'd much rather be able to explicitly opt in to that version, rather than it being done automatically.

rmccue:
Hmm, that is true, forgot about 4.9...

joehoyle:
I updated to "5.2.0" for now, I think that's the better default until we might want to change that decision.

PR #57, opened 2019-05-13
Install sitemeta tables on install

joehoyle:
Otherwise this has to wait for the first admin init, which causes an error on first load.

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -122,7 +122,7 @@ function disable_site_healthcheck_access() {
 }
 
 /**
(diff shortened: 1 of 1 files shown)


inc/namespace.php
```diff
@@ -113,3 +120,14 @@ function disable_site_healthcheck_access() {
 
 	wp_die( 'Site Health not accessible.' );
 }
+
+/**
+ * When WordPress is installed via WP-CLI, run the user-signups setup
```
roborourke:
```suggestion
 * When WordPress is installed via WP-CLI, run the user-signups setup.
```

PR #77, opened 2019-05-24
Strip the default /blog prefix from permalinks

### LINKED ISSUE #70: Remove /blog from default permalinks
This is an annoying multisite weirdness where the `with_front` option is set to true by default. It's unnecessary.

roborourke:
Fixes #70

### DIFF
--- a/inc/permalinks/namespace.php
+++ b/inc/permalinks/namespace.php
@@ -2,6 +2,9 @@
 
 namespace Altis\CMS\Permalinks;
 
+/**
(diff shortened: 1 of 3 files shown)


inc/permalinks/namespace.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+namespace Altis\CMS\Permalinks;
+
+function bootstrap() {
```
rmccue: No function documentation make @rmccue something something.

PR #83, opened 2019-06-13
9: Remove emojis by default

### LINKED ISSUE #9: Plugin: Disable Emojis
@roborourke commented on [Tue Aug 14 2018](https://github.com/humanmade/platform-dev/issues/248)

It's a basic plugin that could be on by default, used on a bunch of internal projects. Basically removes default WP emoji fallbacks to images.

https://wordpress.org/plugins/disable-emojis/

---

@roborourke commented on [Wed Mar 20 2019](https://github.com/humanmade/platform-dev/issues/248#issuecomment-474889985)

Wouldn't use the plugin but we can add this to the CMS module. Moving there.

bradp:
Closes #9

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -52,34 +52,43 @@ function bootstrap() {
 	// Don't show the welcome panel.
 	add_filter( 'get_user_metadata', __NAMESPACE__ . '\\hide_welcome_panel', 10, 3 );
 
-	if ( $config['remove-emojis'] ) {
-		remove_action( 'wp_head', 'print_emoji_detection_script', 7 );
-		remove_action( 'admin_print_scripts', 'print_emoji_detection_script' );
-		remove_action( 'wp_print_styles', 'print_emoji_styles' );
-		remove_action( 'admin_print_styles', 'print_emoji_styles' );
-		remove_filter( 'the_content_feed', 'wp_staticize_emoji' );
-		remove_filter( 'comment_text_rss', 'wp_staticize_emoji' );
-		remove_filter( 'wp_mail', 'wp_staticize_emoji_for_email' );
-		add_filter( 'tiny_mce_plugins', 'disable_emojis_tinymce' );
-		add_filter( 'wp_resource_hints', __NAMESPACE__ . '\\disable_emojis_remove_dns_prefetch', 10, 2 );
+	if ( $config['remove-emoji'] ) {
+		remove_emoji();
 	}
+}
+
+/**
+ * Remove all the emoji scripts that are included with WordPress.
+ */
+function remove_emoji() {
+	remove_action( 'wp_head', 'print_emoji_detection_script', 7 );
+	remove_action( 'admin_print_scripts', 'print_emoji_detection_script' );
+	remove_action( 'wp_print_styles', 'print_emoji_styles' );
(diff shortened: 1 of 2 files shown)


inc/namespace.php
```diff
@@ -51,6 +51,38 @@ function bootstrap() {
...
+	if ( $config['remove-emojis'] ) {
+		remove_action( 'wp_head', 'print_emoji_detection_script', 7 );
+		remove_action( 'admin_print_scripts', 'print_emoji_detection_script' );
+		remove_action( 'wp_print_styles', 'print_emoji_styles' );
+		remove_action( 'admin_print_styles', 'print_emoji_styles' );
+		remove_filter( 'the_content_feed', 'wp_staticize_emoji' );
+		remove_filter( 'comment_text_rss', 'wp_staticize_emoji' );
+		remove_filter( 'wp_mail', 'wp_staticize_emoji_for_email' );
+		add_filter( 'tiny_mce_plugins', 'disable_emojis_tinymce' );
+		add_filter( 'wp_resource_hints', __NAMESPACE__ . '\\disable_emojis_remove_dns_prefetch', 10, 2 );
```
rmccue: Can we move these actions/etc into a `remove_emoji` function instead? This function is getting a bit long now.

inc/namespace.php
```diff
@@ -51,6 +51,38 @@ function bootstrap() {
...
+	}
+}
+
+/**
+ * Remove emoji DNS prefetching.
+ *
+ * @param array  $urls          URLs for resources.
+ * @param string $relation_type Relation type.
+ * @return array                New array with resources.
```
rmccue: Param docs shouldn't be aligned

inc/namespace.php
```diff
@@ -51,6 +51,38 @@ function bootstrap() {
...
+ */
+function disable_emojis_remove_dns_prefetch( $urls, $relation_type ) {
```
rmccue: Typehints for the params and return values please!

inc/namespace.php
```diff
@@ -51,6 +51,38 @@ function bootstrap() {
...
+	if ( 'dns-prefetch' == $relation_type ) {
```
rmccue: This can flipped to an early return instead.

inc/namespace.php
```diff
@@ -51,6 +51,38 @@ function bootstrap() {
...
+		// Strip out any URLs referencing the WordPress.org emoji location
+		$emoji_svg_url_bit = 'https://s.w.org/images/core/emoji/';
+		foreach ( $urls as $key => $url ) {
+			if ( strpos( $url, $emoji_svg_url_bit ) !== false ) {
+				unset( $urls[$key] );
```
rmccue: Lint error.

load.php
```diff
@@ -23,6 +23,7 @@
 		'login-logo'    => '/vendor/altis/cms/assets/logo.svg',
 		'shared-blocks' => true,
 		'default-theme' => 'base',
+		'remove-emojis' => true,
```
rmccue: Can we rename to `remove-emoji`? 😬

roborourke: classic Ryan :)

inc/namespace.php
roborourke: Should this be a strict comparison?

PR #93, opened 2019-09-03
Override the Add Site network admin page

### LINKED ISSUE #50: Show both subdomain and subdirectory UI on new site page
WordPress's Add Site screen changes depending on what the `SUBDOMAIN_INSTALL` constant is set to (see #69). This is an artefact of when this used to matter in WordPress, but one that we can safely rework.

We should install allow users to create subdomain, subdirectory, and custom domain sites as they wish. We might want to have some direction to point them towards the best practices in the process though.

joehoyle:
The default WordPress multisite add site page is lacking in a couple ways. It only allows a subdir / subdomain to be chosed, rather than arbitrary URL (which WordPress supports without a problem), and it also asks for an admin_email, which is mostly not needed.

Rather than trying to override this part of the admin via hooks, which would be very hacky, replacing the whole page will give us the most flexibility.

Fixes #50.


rmccue:
I think we need to guide users a little more than just giving them a freeform text field; @grarighe any thoughts? I was thinking something a little more like this for the UI:

Screen Shot 2019-09-04 at 13 06 39

For reference, here's a comparison:

WordPress | Joe's PR | My Mockup
--|--|--
Screen Shot 2019-09-04 at 13 07 27 | Screen Shot 2019-09-04 at 13 07 31 | Screen Shot 2019-09-04 at 13 06 39

@joehoyle Any reason you removed the language selector as well?

rmccue:
As an alternative idea, we could have a selector which guides users even further to specific choices, and have the site slug input change depending on which item they've selected:

Screen Shot 2019-09-04 at 13 15 45

roborourke:
@rmccue I think your second mockup is more straightforward as there are too many inputs in the first one and it's a very uncommon pattern to mix up inputs and radio buttons like that.

Would also make labelling the inputs a bit too clumsy from an a11y standpoint.

While I think it's clearer to differentiate between subdomain and subdirectory do we care if someone wants to use both a subdomain and directory? Will they know they can use the custom domain option for that? Does it even matter?

Another configuration approach would be to first choose a custom domain or from a list of existing domains then optionally fill in a subdomain and path.

rmccue:
> While I think it's clearer to differentiate between subdomain and subdirectory do we care if someone wants to use both a subdomain and directory?

I *think* WordPress's routing does differentiate a bit, so I think we probably want to reflect the underlying truth here.

> Another configuration approach would be to first choose a custom domain or from a list of existing domains then optionally fill in a subdomain and path.

That could work too, provided it scales.

inc/add_site_ui/namespace.php
```diff
@@ -0,0 +1,133 @@
...
+/**
+ * Register the menu page for the overrided Add New Site page.
+ *
+ * @return void
+ */
+function register_menu_page() {
+	if ( ! is_network_admin() ) {
+		return;
+	}
+	remove_submenu_page( 'sites.php', 'site-new.php' );
```
rmccue: I wonder whether we should hook in similar to the About page so that the URL doesn't change? This would ensure existing links work.

If not, we should at least redirect.

joeleenk:
@roborourke I think this is pretty close to complete now. I still feel like I need to do more url validation, but I'd like you to take a look when you have a chance. Thanks!

roborourke:
@joeleenk sorry for the delay, I'll get this reviewed tomorrow

inc/add_site_ui/namespace.php
```diff
@@ -0,0 +1,291 @@
...
+	$public      = sanitize_text_field( $_POST['public'] );
+	$domain      = '';
+	$path        = '';
+
+	if ( ( strpos( $url, $network_url ) && strpos( $url, 'http' ) !== 0 )
+		|| ( 'site-custom-domain' === $site_type && strpos( $url, 'http' ) !== 0 ) ) {
+		$url = 'https://' . $url;
+	}
+
+	if ( wp_http_validate_url( $url ) ) {
```
roborourke: We'll need to account for this in the Local Server module as it rejects local IP addresses. Just making a note here.

inc/add_site_ui/namespace.php
```diff
@@ -0,0 +1,291 @@
...
+		$url_array = wp_parse_url( $url );
+		$domain    = str_replace( $network_url, '', $url_array['host'] );
+		$domain    = trim( $domain, '.' );
+		$path      = trim( $url_array['path'], '/' );
+	} else {
+		$domain = trim( $url, '.' );
+		$path   = trim( $url, '/' );
```
roborourke: What was the reasoning for using `$url` to set both of these values? I think there's some mixed logic here because `wp_http_validate_url()` will not always be true for a custom domain, there are many reasons why it can fail.

We should ensure that these values are only set under the right `$site_type` contexts.

roborourke:
I pushed up some updates to resolve the logic around `wp_http_validate_url()` as I know you're busy elsewhere @joeleenk - I was able to figure out what was happening.

@rmccue @joehoyle would be good to get the once over from either of you, I think it's alright to merge this.

inc/add_site_ui/namespace.php
```diff
@@ -0,0 +1,296 @@
...
+
+			submit_button( __( 'Add Site' ), 'primary', 'add-site' );
+			?>
+		</form>
+	</div>
+	<?php
+
+	require ABSPATH . 'wp-admin/admin-footer.php';
+
+	// Exit before we attempt to render WordPress' about page.
```
rmccue:
```suggestion
	// Exit before we attempt to render WordPress' built-in page.
```

inc/add_site_ui/namespace.php
```diff
@@ -0,0 +1,296 @@
...
+	exit;
+}
+
+/**
+ * Handler function for the Add Site form submission.
+ */
+function add_site_form_handler() {
```
rmccue: This is a very long function, can we split it up into one function per type?

inc/add_site_ui/namespace.php
```diff
@@ -0,0 +1,296 @@
...
+
+	$network_url = wp_parse_url( network_site_url() );
+	$network_url = $network_url['host'];
+	$form_url    = sanitize_text_field( $_POST['url'] );
+	$url         = $form_url;
+	$title       = sanitize_text_field( $_POST['title'] );
+	$language    = sanitize_text_field( $_POST['language'] ) ?? '';
+	$public      = sanitize_text_field( $_POST['public'] );
+	$domain      = '';
+	$path        = '';
```
rmccue: These lines shouldn't be aligned.

inc/branding/namespace.php
```diff
@@ -77,9 +77,16 @@ function add_color_scheme() {
...
+ *
+ * @param string $hook
  */
-function enqueue_admin_scripts() {
+function enqueue_admin_scripts( string $hook ) {
+
 	wp_enqueue_style( 'altis-branding', plugin_dir_url( dirname( __FILE__, 2 ) ) . 'assets/branding.css', [], '2019-04-24-1' );
+
+	if ( $hook === 'site-new.php' ) {
+		wp_enqueue_script( 'customize-settings', plugin_dir_url( dirname( __FILE__, 2 ) ) . 'assets/customize-settings.js', [], false, true );
```
rmccue:
```suggestion
		wp_enqueue_script( 'altis-customize-settings', plugin_dir_url( dirname( __FILE__, 2 ) ) . 'assets/customize-settings.js', [], false, true );
```

rmccue:
Looking slick.
altis-v3-add-site

Few things I noticed:

Custom domain could use more hints I think; maybe a placeholder like `example.com`

Custom domains don't work properly; `wp_http_validate_url` internally calls `gethostbyname` to determine if it's a loopback request, but that "returns the IPv4 address **or a string containing the unmodified hostname on failure**" which `wp_http_validate_url()` treats as invalid. There's [some code from Happytables we could reuse here](https://github.com/humanmade/happytables/blob/dcb0d5d27049532bee3ff416a5e88eee5bdcfb7c/content/plugins-mu/ht-domains/ht-domains.php#L88-L128) (although it ultimately verifies the domain by requiring DNS to resolve first).

For consistency, "Uncheck to require login." should probably be on a newline (and `.description`) instead.

rmccue:
Working on fixing the above issues and some others on a new branch.

inc/branding/namespace.php
rmccue: Actually also, this needs to be moved across to the right file, otherwise it'll break when branding is disabled.

rmccue:
Changes submitted in #128; that PR should be merged first, and it will be merged into this PR.

roborourke:
@rmccue yep yep yep

Issue #10, opened 2019-03-22
Disable recovery mode


rmccue:
See [the Make/Core post](https://make.wordpress.org/core/2019/03/20/the-improved-fatal-error-protection/). We should disable this, as it's a potential attack vector, and we don't want the ability to trigger the recovery mode. Fatal errors should not ship to production in the first place, and if they do, they should be fixed rather than the entire plugin being disabled.

pdewouters:
This is already disabled https://github.com/humanmade/altis-cms/blob/master/inc/namespace.php#L42

PR #108, opened 2019-10-09
Disable recovery mode

pdewouters:
See https://github.com/humanmade/altis-cms/issues/10

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -55,6 +55,10 @@ function bootstrap() {
 	if ( $config['remove-emoji'] ) {
 		add_action( 'plugins_loaded', __NAMESPACE__ . '\\remove_emoji' );
 	}
+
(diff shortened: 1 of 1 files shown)


pdewouters:
Actually, when I trigger an error `trigger_error("Fatal error", E_USER_ERROR);` in the code, I just see the default browser error page, the recovery mode doesn't seem to kick in. ( this is without disabling it explicitely. Do we have any other code that listens for fatal errors?

The other option would be to override the class

> The entire shutdown handler can be overridden by using a fatal-error-handler.php drop-in. This drop-in must return an instance of a class extending the default WP_Fatal_Error_Handler class. If that is the case, the instance will be used instead of the default one.

pdewouters:
Actually this has already been implemented! https://github.com/humanmade/altis-cms/blob/master/inc/namespace.php#L42

PR #109, opened 2019-10-10
Add option to disable XML RPC

pdewouters:


inc/namespace.php
```diff
@@ -29,6 +29,12 @@ function bootstrap() {
 		Block_Editor\bootstrap();
 	}
 
+	if ( $config['xmlrpc'] === false ) {
+		add_filter( 'xmlrpc_enabled', __NAMESPACE__ . '\\filter_xmlrpc_handler', 999 );
```
roborourke: You could use `__return_false` as the callback here I think, or did you find that those functions aren't loaded at this stage in an XMLRPC request?

inc/namespace.php
```diff
@@ -29,6 +29,12 @@ function bootstrap() {
...
+		add_filter( 'xmlrpc_methods', __NAMESPACE__ . '\\filter_xmlrpc_methods_handler', 999 );
```
roborourke: Similar here too - there's a `__empty_array` callback provided with WP out of the box for this sort of scenario.

inc/namespace.php
```diff
@@ -29,6 +29,12 @@ function bootstrap() {
...
+		add_filter( 'xmlrpc_element_limit', __NAMESPACE__ . '\\filter_xmlrpc_element_limit_handler', 999 );
```
roborourke: I'm curious why all of these filters are needed and not just the `xmlrpc_enabled` one?

inc/namespace.php
pdewouters: I was going to do that - but I saw this code https://github.com/humanmade/altis-cms/blob/master/inc/namespace.php#L139 , so thought it was a convention or for testing purposes?

inc/namespace.php
pdewouters: There's a comment in core:
```
 		 * Contrary to the way it's named, this filter does not control whether XML-RPC is *fully*
		 * enabled, rather, it only controls whether XML-RPC methods requiring authentication - such
		 * as for publishing purposes - are enabled.
```

inc/namespace.php
roborourke: @pdewouters that makes sense - so in the case you linked to it's because WP hasn't fully loaded by the time that filter is run. In the case of the XMLRPC those functions are be available.

inc/namespace.php
roborourke: Wow well that's fun 🤦‍♂ 
I did spot a potential nuclear option too as well if we want to go whole hog:

```php
$wp_xmlrpc_server_class = apply_filters( 'wp_xmlrpc_server_class', 'wp_xmlrpc_server' );
$wp_xmlrpc_server       = new $wp_xmlrpc_server_class;

// Fire off the request
$wp_xmlrpc_server->serve_request();
```

We could add a fake XMLRPC class that only has an empty `serve_request()` method that either does nothing or returns a 404 page.

pdewouters: @rmccue would this option be preferable?

roborourke: For reference I think what you have here with these three filters is fine right now, I jjust found it interesting. The above suggestion would require a bunch of testing as it could affect compatibility with some plugins.

roborourke:
@pdewouters I deleted the `.gitignore` file, shouldn't have been part of this repo as it's the project root `.gitignore`. Last job is to document this then, copy a page like `branding.md` and update the copy to explain a bit about XMLRPC and then how to disable it through the config.

docs/xmlrpc.md
```diff
@@ -0,0 +1,21 @@
+# XML RPC
+
+WordPress XML RPC is a remote procedure call which uses XML to encode its calls and HTTP as a transport mechanism.
```
roborourke:
```suggestion 
XML RPC (Remote Procedure Call) is a type of API which uses XML to encode its calls and HTTP as a transport mechanism.
```

docs/xmlrpc.md
```diff
@@ -0,0 +1,21 @@
...
+
+It is used by certain applications to enable remote management of WordPress sites, for example certain mobile apps for moderating comments, or the Jetpack plugin.
```
roborourke:
```suggestion
It is used to enable remote management of applications, for example via mobile or desktop apps.
```
Let's not mention Jetpack :)

docs/xmlrpc.md
```diff
@@ -0,0 +1,21 @@
...
+
+By default the XML RPC is enabled. If you need to disable it, set the `xmlrpc` configuration property to `false`
```
roborourke:
```suggestion
By default XML RPC is enabled. If you wish to disable it, set the `xmlrpc` configuration property to `false`
```

PR #110, opened 2019-10-11
WIP Add object cache health check

pdewouters:
This is a work in progress


roborourke:
Going to close this out as it's now stale, and we have an object cache healthcheck there. Need to more generally revisit the user facing healthcheck stuff that's part of WP in future.

PR #118, opened 2019-10-14
Add bcrypt password plugin

pdewouters:


composer.json
```diff
@@ -11,6 +11,7 @@
   "require": {
     "php": ">=7.1",
     "stuttter/wp-user-signups": "3.1.0",
+    "roots/wp-password-bcrypt": "^1.0",
```
roborourke: We don't control this plugin so lets add it as a fixed version number instead:
```suggestion
    "roots/wp-password-bcrypt": "1.0.0",
```

PR #128, opened 2019-11-16
Improve and fix Add Site errors

rmccue:
* Refactors the form handler into something more readable
* Adds validation for domains and paths (reserved paths now blocked, domains validated properly)
* Fixes up some JS problems
* Fixes issue with branding CSS
* Adds placeholders to indicate valid inputs

altis-v3-add-site

Built off #93.

### DIFF
--- a/assets/customize-settings.js
+++ b/assets/customize-settings.js
@@ -11,11 +11,6 @@ const siteSubdomain = document.getElementById( 'site-subdomain' ),
 let domainTextEl = document.createElement( 'span' );
 domainTextEl.className = 'site-address-domain-text';
 
-// Add an event listener to each radio button.
-domainTypeRadioBtns.forEach( function ( element ) {
-	element.addEventListener( 'change', updatedomainText );
-} );
-
 // Add domain hints around the site address field.
 function updatedomainText( event ) {
 	switch ( event.target ) {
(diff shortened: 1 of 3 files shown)


assets/customize-settings.js
```diff
@@ -1,43 +1,52 @@
...
+( function () {
+	/*
+	* Custom JS to modify the Altis CMS module Add New Site page
+	*/
+	const siteSubdomain = document.getElementById( 'site-subdomain' );
+	const siteSubdirectory = document.getElementById( 'site-subdirectory' );
+	const siteCustomDomain = document.getElementById( 'site-custom-domain' );
+	const siteAddress = document.getElementById( 'site-address' );
+	const siteAddressDesc = document.getElementById( 'site-address-desc' );
+	const networkAddress = window.location.hostname;
```
roborourke: Can the network admin be accessed via subsite or always on the primary hostname?

inc/branding/namespace.php
```diff
@@ -17,7 +17,7 @@
  */
 function bootstrap() {
 	add_action( 'add_admin_bar_menus', __NAMESPACE__ . '\\remove_wordpress_admin_bar_item' );
-	add_action( 'admin_enqueue_scripts', __NAMESPACE__ . '\\enqueue_admin_scripts' );
+	add_action( 'admin_bar_init', __NAMESPACE__ . '\\enqueue_admin_scripts' );
```
roborourke: 😱 picking up my bad habits!

roborourke: Also the admin bar can be turned off

assets/customize-settings.js
rmccue: Yes, but I think this matches what WordPress already does? In any case, I didn't change this code, just refactored.

inc/branding/namespace.php
rmccue: This is actually intentionally like this, as it needs to enqueue on the frontend as well. The change in here broke the Altis logo in the admin bar on the frontend.

PR #133, opened 2019-11-27
Fix MULTISITE handling during install

roborourke:
There are 2 additions here:

- prevent web access to the install screen
- define a constant if we detect the cli command is for the initial install

Relies on an update to CMS Installer.


inc/cli/namespace.php
```diff
@@ -0,0 +1,37 @@
+<?php
+
+namespace Altis\CMS\CLI;
+
+use WP_CLI;
+
+function bootstrap() {
```
rmccue: phpdoc? :)

inc/namespace.php
```diff
@@ -222,3 +226,17 @@ function hide_welcome_panel( $value, int $user_id, string $meta_key ) {
...
+ */
+function prevent_web_install() {
+	if ( $_SERVER['REQUEST_URI'] !== '/wp-admin/install.php' ) {
+		return;
+	}
+
+	// Return 200 status for healthcheck.
+	status_header( 200 );
+	echo 'This site is currently unavailable';
+	exit;
```
rmccue: This seems familiar somehow 🤔

inc/namespace.php
```diff
@@ -222,3 +226,17 @@ function hide_welcome_panel( $value, int $user_id, string $meta_key ) {
...
+/**
+ * Prevent direct web access to wp-admin/install.php.
+ */
+function prevent_web_install() {
+	if ( $_SERVER['REQUEST_URI'] !== '/wp-admin/install.php' ) {
+		return;
+	}
+
+	// Return 200 status for healthcheck.
+	status_header( 200 );
```
rmccue: This is actually not necessary, but let's keep it anyway so it's explicit 👍

inc/namespace.php
```diff
@@ -222,3 +226,17 @@ function hide_welcome_panel( $value, int $user_id, string $meta_key ) {
 
 	return [ 0 ];
 }
+
+/**
+ * Prevent direct web access to wp-admin/install.php.
+ */
+function prevent_web_install() {
+	if ( $_SERVER['REQUEST_URI'] !== '/wp-admin/install.php' ) {
```
rmccue: I guess we probably can't detect this a different way? Like, some sort of `WP_INSTALLING` constant?

roborourke: Not that I can see

PR #141, opened 2019-12-05
#123 Remove auto updates for core

### LINKED ISSUE #123: Set WP_AUTO_UPDATE_CORE to false
We've removed the updates capability but not set the related constants which leads to failed attempts and error logs.

kirstyburgoine:
Uses `auto_update_core` filter

Fixes #123

### DIFF
--- a/inc/remove_updates/namespace.php
+++ b/inc/remove_updates/namespace.php
@@ -8,6 +8,7 @@ namespace Altis\CMS\Remove_Updates;
 function bootstrap() {
 	add_action( 'admin_init', __NAMESPACE__ . '\\remove_maybe_update_checks', 1 );
 	add_action( 'admin_init', __NAMESPACE__ . '\\remove_update_nag' );
+	add_action( 'admin_init', __NAMESPACE__ . '\\remove_auto_update' );
 	add_action( 'plugins_loaded', __NAMESPACE__ . '\\remove_update_check_cron' );
 	add_filter( 'map_meta_cap', __NAMESPACE__ . '\\remove_update_capabilities', 10, 2 );
 }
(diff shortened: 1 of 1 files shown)


roborourke:
@kirstyburgoine I can't seem to find where this filter is used in WP, searching the codebase just takes me to the `WP_AUTO_UPDATE_CORE` constant

kirstyburgoine:
@roborourke Sorry, this should have had WIP on it still. I found the filter here: https://wordpress.org/support/article/configuring-automatic-background-updates/#core-updates-via-filter but I haven't tested locally yet

roborourke:
Hm, I think those docs are wrong or maybe for a future update :/ I can see the specific update type filters but not that global one. The constant is definitely present though

roborourke:
Lets define this as false too `AUTOMATIC_UPDATER_DISABLED`

kirstyburgoine:
@roborourke I've updated this to use `AUTOMATIC_UPDATER_DISABLED` now, I haven't also added `WP_AUTO_UPDATE_CORE` as per the original ticket because according to this doc: https://wordpress.org/support/article/configuring-automatic-background-updates/ `AUTOMATIC_UPDATER_DISABLED` will automatically disable all updates including core updates so it doesn't seem to be needed

inc/remove_updates/namespace.php
```diff
@@ -8,6 +8,7 @@
 function bootstrap() {
 	add_action( 'admin_init', __NAMESPACE__ . '\\remove_maybe_update_checks', 1 );
 	add_action( 'admin_init', __NAMESPACE__ . '\\remove_update_nag' );
+	add_action( 'admin_init', __NAMESPACE__ . '\\remove_auto_updates' );
```
roborourke: Last thing to do here - I don't think `admin_init` will always be fired in the context of upgrades / installs so you can call the function directly from here, it's fine as we're just defining a constant so earlier the better.

inc/remove_updates/namespace.php
```diff
@@ -38,6 +39,14 @@ function remove_update_nag() {
 	remove_filter( 'update_footer', 'core_update_footer' );
 }
 
+/**
+ * Remove all automatic updates.
+ *
```
roborourke:
```suggestion
```

PR #156, opened 2019-12-10
[Backport to v2-branch]: Add WP_INITIAL_INSTALL constant

roborourke:
Backporting #133


composer.json
```diff
@@ -12,12 +12,14 @@
     "php": ">=7.1",
     "stuttter/wp-user-signups": "3.1.0",
     "johnpbloch/wordpress": "5.2.4",
-    "altis/cms-installer": "~0.2.1"
+    "altis/cms-installer": "^0.4.0"
```
rmccue: Is this going to break compat? Presumably we'll have those Composer installation errors, but this will now happen on a minor version update as well, which isn't great

roborourke: It'd require an update to the CMS installer package either way unfortunately due to the `wp-config.php` modification. It def isn't great so could just do it for the RC branch if you think that's more acceptable. Alternatively just get v3 on the auth stack early.

PR #171, opened 2020-01-08
Fixing undefined variable notice with custom domain

### LINKED ISSUE #167: Add Site UI requires logic checks for URL parsing,
Testing the Add site UI locally, I entered `peterwilsonme.local/~` as a custom domain, as I neglected the scheme a few undefined variable and index notification were thrown.

peterwilsoncc_local_wp-admin_network_site-new_php

kevinlangleyjr:
Fixes undefined variable notice when including the schema for a with custom domain.

Fixes #167

### DIFF
--- a/inc/add_site_ui/namespace.php
+++ b/inc/add_site_ui/namespace.php
@@ -418,11 +418,11 @@ function handle_subdirectory( string $value ) : ?array {
  * Handle a custom domain input value.
  *
  * @param string $url Input custom domain value.
- * @return array|null $value Assoc array with domain and path keys, or null if invalid.
+ * @return array|null $url Assoc array with domain and path keys, or null if invalid.
  */
 function handle_custom_domain( string $url ) : ?array {
-	if ( strpos( $value, 'http' ) !== 0 ) {
-		$url = 'https://' . $value;
+	if ( strpos( $url, 'http' ) !== 0 ) {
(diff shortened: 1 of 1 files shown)


inc/add_site_ui/namespace.php
```diff
@@ -417,10 +417,10 @@ function handle_subdirectory( string $value ) : ?array {
...
  *
- * @param string $value Input custom domain value.
+ * @param string $url Input custom domain value.
  * @return array|null $value Assoc array with domain and path keys, or null if invalid.
  */
-function handle_custom_domain( string $value ) : ?array {
+function handle_custom_domain( string $url ) : ?array {
 	if ( strpos( $value, 'http' ) !== 0 ) {
 		$url = 'https://' . $value;
 	}
```
roborourke: The issue appears to be this bit of logic rather than the argument name - this change as it is will cause the same problem with `$value` being undefined. I think either this needs to change the `$value` variables to `$url` or have an `else` block that just contains `$url = $value;`

kevinlangleyjr: Good point, sorry I missed that! I think if we just change the `$value` variables to `$url` we get the same outcome but without needing an `else` block that just sets the `$url` equal to `$value`. So, I did just that in my latest commit.

Issue #115, opened 2019-10-11
Altis Admin Bar Frontend Spacing


tomjn:
On the frontend the Altis admin bar logo has full WP Admin spacing, despite there being no admin bar:

<img width="438" alt="Screenshot 2019-10-11 at 16 34 58" src="https://user-images.githubusercontent.com/58855/66664772-16260600-ec45-11e9-99c0-161ea13e2f72.png">

It looks as if an item is present but hidden or missing

tomjn:
First thoughts on wrangling this would be to have the CSS work from a collapsed by default viewpoint, then override for when the admin body class is present, and again for folded, but there's probably a simpler set of rules to do it

kevinlangleyjr:
Should we only be targeting the frontend with this styling change?

I personally think even in the backend it looks better to have the rest of the admin bar right next to the Altis logo.

**Current**
![Screen Shot 2020-01-09 at 1.10.02 PM.png](https://images.zenhubusercontent.com/544ec2c47c6ccc6277a58b71/bb6f3787-bed0-4a50-a218-b75b36854923)

**Proposed**
![Screen Shot 2020-01-09 at 1.09.25 PM.png](https://images.zenhubusercontent.com/544ec2c47c6ccc6277a58b71/ad6f788d-5fdd-447c-9c34-9f9ecb6e2b75)

sambulance:
PR: #175

roborourke:
It was an explicit decision to do that yeah. Thanks for fixing it on the front end Sam!

PR #175, opened 2020-01-10
#115 Fix Altis spacing in the admin bar

sambulance:
Sets the width of the Altis menu item in the menu bar in the admin only, and when the sidebar is not collapsed (folded). 

Frontend
Screenshot 2020-01-10 at 12 27 39

Admin
Screenshot 2020-01-10 at 12 27 59

Admin - sidebar collapsed
Screenshot 2020-01-10 at 12 28 14

See #115


mattheu:
This gap has always made me feel a little uneasy! Thanks for fixing it!

PR #179, opened 2020-01-25
Replace WP preview animation with Altis animation

kevinlangleyjr:
Fixes: https://github.com/humanmade/platform-dev/issues/381

Uses pure CSS to animate the SVG instead of JS like the original ticket requested.

### DIFF
--- a/inc/branding/namespace.php
+++ b/inc/branding/namespace.php
@@ -313,7 +313,7 @@ function enqueue_block_editor_branding_assets() {
 			<svg version="1.1" id="altis-svg-shape" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px" viewBox="0 0 243 226.8" xml:space="preserve">
 				<path class="altis-shape" d="M22.05,38.81,220.44,3.86C218.63,8.77,159.21,169.94,155.18,181c-4.65,12.8-12.37,27.8-28.61,25.79A24.35,24.35,0,0,1,108,195.67a25.23,25.23,0,0,1-3.44-9.14c-2.78-15.74-19-97.66-61.12-134.4,0,0-12.11-10.52-21.4-13.32a40.15,40.15,0,0,0-7.84,3.09C-4.91,53.05,3.3,77.11,17.49,86,32,95,90.11,96.73,133.73,102.58s40.34,27.19,40.34,27.19l-2.51,6.8c-3.27,7.54-20.31,15.69-29.16-2.47S132.06,53.44,87.69,27.24"/>
 			</svg>
-			<div class="loading">' . __( 'Loading the future...', 'altis' ) . '</div>
+			<div class="loading">' . esc_html__( 'Loading the future...', 'altis' ) . '</div>
 		</div>
 		<style>
 			body {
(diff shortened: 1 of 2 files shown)


inc/branding/namespace.php
```diff
@@ -290,3 +291,92 @@ function remove_howdy_greeting( WP_Admin_Bar $wp_admin_bar ) {
...
+		false,
+		true
+	);
+
+	$markup = '
+		<div class="editor-post-preview-button__interstitial-message">
+			<svg version="1.1" id="altis-svg-shape" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px" viewBox="0 0 243 226.8" xml:space="preserve">
+				<path class="altis-shape" d="M22.05,38.81,220.44,3.86C218.63,8.77,159.21,169.94,155.18,181c-4.65,12.8-12.37,27.8-28.61,25.79A24.35,24.35,0,0,1,108,195.67a25.23,25.23,0,0,1-3.44-9.14c-2.78-15.74-19-97.66-61.12-134.4,0,0-12.11-10.52-21.4-13.32a40.15,40.15,0,0,0-7.84,3.09C-4.91,53.05,3.3,77.11,17.49,86,32,95,90.11,96.73,133.73,102.58s40.34,27.19,40.34,27.19l-2.51,6.8c-3.27,7.54-20.31,15.69-29.16-2.47S132.06,53.44,87.69,27.24"/>
+			</svg>
+			<div class="loading">Loading the future...</div>
```
roborourke: 😂

roborourke: I think this needs to be translatable as it's visible to users

kevinlangleyjr: Whoops, can't believe I missed that. Should be good now!

inc/branding/namespace.php
```diff
@@ -290,3 +291,92 @@ function remove_howdy_greeting( WP_Admin_Bar $wp_admin_bar ) {
...
+		false,
+		true
+	);
+
+	$markup = '
+		<div class="editor-post-preview-button__interstitial-message">
+			<svg version="1.1" id="altis-svg-shape" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px" viewBox="0 0 243 226.8" xml:space="preserve">
+				<path class="altis-shape" d="M22.05,38.81,220.44,3.86C218.63,8.77,159.21,169.94,155.18,181c-4.65,12.8-12.37,27.8-28.61,25.79A24.35,24.35,0,0,1,108,195.67a25.23,25.23,0,0,1-3.44-9.14c-2.78-15.74-19-97.66-61.12-134.4,0,0-12.11-10.52-21.4-13.32a40.15,40.15,0,0,0-7.84,3.09C-4.91,53.05,3.3,77.11,17.49,86,32,95,90.11,96.73,133.73,102.58s40.34,27.19,40.34,27.19l-2.51,6.8c-3.27,7.54-20.31,15.69-29.16-2.47S132.06,53.44,87.69,27.24"/>
+			</svg>
+			<div class="loading">' . __( 'Loading the future...', 'altis' ) . '</div>
```
roborourke: I just checked if this would execute an inline script and it will, so best to use `esc_html__()` here. Yay Gutenberg :)

roborourke:
Thanks @kevinlangleyjr & @sambulance! Looks ace

Issue #104, opened 2019-10-02
Add performance enhancements for WP


roborourke:
Core WP has some performance issues that may only reveal themselves in certain situations. We can start adding some simple mitigations in this module:

- force limit comments per page to 50 maximum
- add https://github.com/Automattic/advanced-post-cache
- add https://github.com/Automattic/lightweight-term-count-update

ht/ @johnbillion 

Acceptance criteria:

- [x] Comment pagination is always enabled
- [x] Impossible set comments per page to a value higher than 50
- [ ] Advanced post cache is required in `composer,json`
- [ ] Advanced post cache file is loaded on the `muplugins_loaded` hook in `inc/namespace.php`
- [ ] Lightweight term count update is required in `composer.json`
- [ ] Lightweight term count is loaded on the `muplugins_loaded` hook in `inc/namespace.php`
- [ ] Above plugins are added to the exclude list in humanmade/altis-core here https://github.com/humanmade/altis-core/blob/master/inc/composer/class-installer.php#L26
- [ ] Review and document performance tweaks from https://github.com/Automattic/vip-go-mu-plugins-built/tree/master/performance here to see what else we could add

johnbillion:
Background:

* Advanced Post Cache caches post queries in regular loops. I've not tested it heavily but it does remove the main post query once the cache is warm.
* Lightweight Term Count Update removes the slow queries that core performs when recalculating the number of posts assigned to terms after updating a post's terms. On the RS DMS project with ~6M rows in the posts table these queries take 20 full seconds for each taxonomy. Notably, this isn't enabled during cron and CLI requests, which is where we were seeing terrible performance problems recently. We enabled it for cron events according to https://github.com/Automattic/lightweight-term-count-update/issues/9.

In addition:

* VIP implements a meta_key+meta_value index on the postmeta table. The index can get 
huge but it does allow for performant meta_value queries, and we're actively using them on RS DMS. https://github.com/Automattic/vip-go-mu-plugins-built/blob/97ee6f2abc79bd1bb5e9149f5d63e5f201d179be/schema.php . Joe H indicated he'd be hesitant about implementing this as-is, but it's certainly something to consider if for some reason ES isn't a viable solution for a particular site.

johnbillion:
VIP has some more performance tweaks here but I've not looked at them in depth: https://github.com/Automattic/vip-go-mu-plugins-built/tree/master/performance

dashaluna:
[...]

PR #180, opened 2020-01-27
Force comment pagination & limit to max 50

dashaluna:
**NB** This PR is for comments part only of the main ticket #104

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -71,7 +71,7 @@ function bootstrap() {
 	add_filter( 'admin_email_check_interval', '__return_zero' );
 
 	/*
-	 * Performance enhancements for comments as best practise.
+	 * Performance enhancements for comments as best practice.
 	 * This is to prevent the CMS generating large HTML pages
 	 * in cases where there may be 1000s of comments,
 	 * as commenters will not see a cached version of the page.
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -268,7 +268,7 @@ function prevent_web_install() {
  *
  * @return int Number of comments per page.
  */
-function set_comments_per_page( $value ) {
+function set_comments_per_page( $value ) : int {
 	$value = intval( $value );
 	return $value <= 50 ? $value : 50;
 }
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -257,7 +257,7 @@ function prevent_web_install() {
 }
 
 /**
- * Set comments per page to be 50 max as best practise.
+ * Set comments per page to be 50 max as best practice.
  *
  * This is to prevent the CMS generating large HTML pages
  * in cases where there may be 1000s of comments,
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -70,15 +70,17 @@ function bootstrap() {
 	// Disable the admin_email verification interval.
 	add_filter( 'admin_email_check_interval', '__return_zero' );
 
-	// Performance enhancements.
+	/*
+	 * Performance enhancements for comments as best practise.
+	 * This is to prevent the CMS generating large HTML pages
+	 * in cases where there may be 1000s of comments,
+	 * as commenters will not see a cached version of the page.
+	 */
 	// Comment pagination is always enabled.
-	add_filter( 'pre_option_page_comments', '__return_true', 999 );
+	add_filter( 'pre_option_page_comments', '__return_true' );
 
 	// Force limit comments per page to 50 max.
-	add_filter( 'pre_update_option_comments_per_page', function( $value ) {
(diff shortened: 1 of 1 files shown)


inc/namespace.php
```diff
@@ -69,6 +69,16 @@ function bootstrap() {
 
 	// Disable the admin_email verification interval.
 	add_filter( 'admin_email_check_interval', '__return_zero' );
+
+	// Performance enhancements.
+	// Comment pagination is always enabled.
+	add_filter( 'pre_option_page_comments', '__return_true', 999 );
+
```
dashaluna: @roborourke I got couple of questions about this:

1. Is it ok to use `999` priority or there is another way to make sure that nothing else can modify this? If someone else adds a filter with grater priority they can change the value.

2. Is it sufficient to use short-circuit filter like this? Or does I have to actually store the new value in the DB as well, so if WP accesses that value directly in DB it's set correctly? I couldn't figure out what the best hook/action to use to do that.

Thanks!

roborourke: @dashaluna 

1. I think we can leave the priority as the default 10. If someone is filtering that value then it's by design and we shouldn't get in the way of that. This is really to protect site admins / users from setting things up badly.

2. Yes this is sufficient, it's very uncommon for the data to be accessed directly so not something I'd worry about. For the parts of core that apply these settings they always use `get_option()` so no need to overthink this :)

inc/namespace.php
```diff
@@ -69,6 +69,16 @@ function bootstrap() {
 
 	// Disable the admin_email verification interval.
 	add_filter( 'admin_email_check_interval', '__return_zero' );
+
+	// Performance enhancements.
+	// Comment pagination is always enabled.
```
roborourke: I'd further explain in this comment that "This is to prevent the CMS generating large HTML pages in cases where there may be 1000s of comments, as commenters will not see a cached version of the page"

inc/namespace.php
```diff
@@ -69,6 +69,16 @@ function bootstrap() {
...
+	add_filter( 'pre_option_page_comments', '__return_true', 999 );
+
+	// Force limit comments per page to 50 max.
+	add_filter( 'pre_update_option_comments_per_page', function( $value ) {
```
roborourke: Let's keep this consistent with the other uses of `add_filter()` in this function and put the function itself at the end of the file. This allows it to be unhooked if desired. We can drop the priority argument here too.

inc/namespace.php
dashaluna: Well the original ticket does say `force limit comments per page to 50 maximum`, so I'm being forceful 💪

dashaluna:
@roborourke thanks for the clear feedback, all makes sense. Fixed things :)

inc/namespace.php
roborourke: Fair point! I think using code is all the "force" we need in this case though. Ultimately you can't really force anything on a developer in WP land.

inc/namespace.php
```diff
@@ -69,6 +69,18 @@ function bootstrap() {
 
 	// Disable the admin_email verification interval.
 	add_filter( 'admin_email_check_interval', '__return_zero' );
+
+	/*
+	 * Performance enhancements for comments as best practise.
```
roborourke:
```suggestion
	 * Performance enhancements for comments as best practice.
```

inc/namespace.php
```diff
@@ -243,3 +255,20 @@ function prevent_web_install() {
 	echo 'This site is currently unavailable';
 	exit;
 }
+
+/**
+ * Set comments per page to be 50 max as best practise.
```
roborourke:
```suggestion
 * Set comments per page to be 50 max as best practice.
```

inc/namespace.php
```diff
@@ -243,3 +255,20 @@ function prevent_web_install() {
...
+ * This is to prevent the CMS generating large HTML pages
+ * in cases where there may be 1000s of comments,
+ * as commenters will not see a cached version of the page.
+ *
+ * @param mixed $value Option value for the 'comments_per_page' option,
+ *                     it's set in WP Admin under Settings -> Discussion -> Other comment settings.
+ *
+ * @return int Number of comments per page.
+ */
+function set_comments_per_page( $value ) {
```
roborourke:
```suggestion
function set_comments_per_page( $value ) : int {
```

inc/namespace.php
dashaluna: Doh, and I did google it... just missed that `practise` is a verb! Thanks.

dashaluna:
@roborourke merged, just flagging this so this PR  doesn't get lost as its main issue is only partly done.

PR #183, opened 2020-01-30
Fixing the all items label for the Block Categories.

kevinlangleyjr:
Looks like a copy/pasta issue.


roborourke:
🤦‍♂  yeah, my bad! Thanks for spotting it

PR #187, opened 2020-01-31
[100] rework signups

jazzsequence:
This attempts to remove the "Sign ups" terminology using existing hooks. 

* Sign ups list page moved into the Users menu with the title "Invitations"
* Removes the Sign Ups top level menu
* Hooks into the `admin_title` filter to change "Sign ups" to "Invitations"

**Note:** This does not actually change the title of the page. I was hoping that there would be a WordPress way to change the page title but I wasn't able to get this working in the time I had.


joehoyle:
Closing as stale

PR #190, opened 2020-02-14
Normalise asset URLs to remove relative paths

roborourke:
Some plugins or code can register asset URLs with relative paths. This does work in the browser however the browser security plugin will throw an error.

We can fix the problem globally by doing this.


rmccue:
I feel like we should probably trigger a warning for this.

roborourke:
Could be annoying if it's coming from a project dependency to do that. It's somewhat benign if we're fixing it this way, I know what you mean though.

roborourke:
@rmccue added a warning when `WP_DEBUG` is true as a compromise, should at least keep the logs on production cleaner and should get picked up locally in query monitor.

PR #192, opened 2020-02-26
Use more specific class name, better theme compat

roborourke:
Fixes some img alignment issues in themes that have global styles for images or use the `.icon` CSS class.


rmccue:
Do we want to backport this?

roborourke:
Yeahhh good shout 👍

PR #208, opened 2020-03-24
Add option to disable RSS feeds

### LINKED ISSUE #205: Add configuration option to disable feeds
By default WordPress exposes feeds for public content. This is not obviously exposed anywhere.

We should add a configuration option to disable feeds, and document it, making it clear that this data is exposed. Note that we do not want to change the default at this time.

Acceptance criteria:
* [ ] When I set `modules.core.remove-feeds` to true in the Altis config, feeds should be disabled
* [ ] When I set up a new Altis installation, feeds should be enabled
* [ ] When I view the CMS documentation, I should be able to clearly see how to configure this option and what the default is

ivdimova:
Fixes #205

### DIFF
--- a/docs/rss.md
+++ b/docs/rss.md
@@ -1,6 +1,6 @@
-# RSS Feed
+# RSS Feeds
 
-RSS Feeds (“Rich Site Summary” or “Really Simple Syndication”), Atom or RDF are type of web feeds that provides users and applications with regular updates from a given website. The RSS feed is a structured XML code document. 
+RSS Feeds ("Rich Site Summary" or "Really Simple Syndication"), Atom or RDF are type of web feeds that provides users and applications with regular updates from a given website. The RSS feed is a structured XML code document. 
 
 By default the RSS Feeds are enabled. If you wish to disable it, set the `feeds` configuration property to `false`
 
--- a/docs/rss.md
+++ b/docs/rss.md
@@ -1,0 +1,19 @@
+# RSS Feed
+
+RSS Feed (“Rich Site Summary” or “Really Simple Syndication”) is type of web feed that provides users and applications with regular updates from a given website. The RSS feed is a structured XML code document.
+
+By default RSS Feed is enabled. If you wish to disable it, set the `rss` configuration property to `false`
+
+```json
+{
+	"extra": {
+		"altis": {
+			"modules": {
+				"cms": {
+					"rss": false
+				}
+			}
+		}
+	}
+}
+```
--- a/docs/rss.md
+++ b/docs/rss.md
@@ -1,8 +1,8 @@
 # RSS Feed
 
-RSS Feed (“Rich Site Summary” or “Really Simple Syndication”) is type of web feed that provides users and applications with regular updates from a given website. The RSS feed is a structured XML code document.
+RSS Feeds (“Rich Site Summary” or “Really Simple Syndication”), Atom or RDF are type of web feeds that provides users and applications with regular updates from a given website. The RSS feed is a structured XML code document. 
 
-By default RSS Feed is enabled. If you wish to disable it, set the `rss` configuration property to `false`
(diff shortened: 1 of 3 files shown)


inc/namespace.php
```diff
@@ -95,6 +113,14 @@ function bootstrap() {
 	add_filter( 'style_loader_src', __NAMESPACE__ . '\\real_url_path', -10, 2 );
 }
 
+/**
+ * Remove the RSS feed and redirect to the home page.
+ */
+function disable_feed() {
+	wp_redirect( home_url() );
+	die();
```
rmccue: Could we send a 404 instead? Potentially we could remove the rewrites entirely.

docs/rss.md
```diff
@@ -0,0 +1,19 @@
...
+
+By default RSS Feed is enabled. If you wish to disable it, set the `rss` configuration property to `false`
+
+```json
+{
+	"extra": {
+		"altis": {
+			"modules": {
+				"cms": {
+					"rss": false
```
rmccue: I'd probably call this `feeds` rather than `rss` (and just broadly call it "Feeds") as it's a bit broader than specifically just RSS.

inc/namespace.php
```diff
@@ -46,6 +46,24 @@ function bootstrap() {
...
+
+		// Disable comment feeds.
+		add_action( 'do_feed_rss2_comments', __NAMESPACE__ . '\\disable_feed', 1 );
+		add_action( 'do_feed_atom_comments', __NAMESPACE__ . '\\disable_feed', 1 );
+
+		// Prevent feed links from being inserted in the <head> of the page.
+		add_action( 'feed_links_show_posts_feed', '__return_false', -1 );
+		add_action( 'feed_links_show_comments_feed', '__return_false', -1 );
+		remove_action( 'wp_head', 'feed_links', 2 );
+		remove_action( 'wp_head', 'feed_links_extra', 3 );
```
rmccue: I think these may need to be hooked in later (i.e. `plugins_loaded`), not sure if default-actions.php has been loaded when this runs?

ivdimova: Looks like these are hooked initially on `wp_head`  in `default-filters.php` so this seems to work fine. Why do you think it needs to be hooked later?

inc/namespace.php
ivdimova: I changed it, would this work?

docs/rss.md
ivdimova: Ok, changed to feeds.

inc/namespace.php
rmccue: I *think* this function runs before `default-filters.php` is loaded (on `altis.modules.load` or whatever the event is called), but I could be wrong on that; if it's working for you, then it's probably OK

ivdimova: @rmccue what's the best way to trace this actually?

ivdimova: Hmm, you are right.

ivdimova: @rmccue I can't seem to figure out a work around this, do you have any suggestion or I should just leave the links be?

docs/rss.md
```diff
@@ -0,0 +1,19 @@
+# RSS Feed
```
rmccue, line 1:
```suggestion
# RSS Feeds
```

docs/rss.md
```diff
@@ -0,0 +1,19 @@
...
+
+RSS Feeds (“Rich Site Summary” or “Really Simple Syndication”), Atom or RDF are type of web feeds that provides users and applications with regular updates from a given website. The RSS feed is a structured XML code document.
```
rmccue, line 3: Curly quotes should be handled in rendering not in the source here.

```suggestion
RSS Feeds ("Rich Site Summary" or "Really Simple Syndication"), Atom or RDF are type of web feeds that provides users and applications with regular updates from a given website. The RSS feed is a structured XML code document. 
```

Issue #96, opened 2019-09-10
Document available configuration


rmccue:
Version 2.0 disabled emoji by default with #83, but we never documented it. There's other options we haven't documented properly, so we should document all of these.

sambulance:
Are the options you're referring to the the ones in [load.php](https://github.com/humanmade/altis-cms/blob/master/load.php)?

```php
	$default_settings = [
		'enabled'       => true,
		'branding'      => true,
		'login-logo'    => '/vendor/altis/cms/assets/logo.svg',
		'shared-blocks' => true,
		'default-theme' => 'base',
		'remove-emoji'  => true,
		'xmlrpc'        => true,
	];
```

Where exactly would these be documented?

joehoyle:
Each module has it's own docs, e.g. https://github.com/humanmade/altis-cms/tree/master/docs. You can view them locally in the admin-bar Altis Logo -> Documentation.

Yes those are the options!

sambulance:
In addition, there are a few third-party plugins added to the CMS module that are not documented.

https://github.com/humanmade/altis-cms/blob/master/composer.json#L13

- "stuttter/wp-user-signups"
- "roots/wp-password-bcrypt"
- "johnbillion/extended-cpts"

kevinlangleyjr:
> * "johnbillion/extended-cpts"

`extended-cpts` has documentation now 👍

jennybeaumont:
Blocked by https://github.com/humanmade/altis-cms/issues/260

PR #216, opened 2020-05-08
add documentation for remove emoji

jazzsequence:
* adds documentation for the remove-emoji module

See #96

### DIFF
--- a/docs/remove-emoji.md
+++ b/docs/remove-emoji.md
@@ -1,0 +1,27 @@
+# Remove Emoji
+
+Remove all the emoji scripts that are included with WordPress. By default, this module is active. To allow emojis, toggle the `remove-emoji` module in the Altis config.
+
+```json
+{
+	"extra": {
+		"altis": {
+			"modules": {
+				"cms": {
+					"remove-emoji": false
+				}
+			}
+		}
+	}
+}
+```
+
+## About emoji
+Emoji are the ideograms or smileys 😃 used in electronic messages and Web pages. Originating in Japan on mobile devices, they are now commonly available on devices worldwide, ranging from mobile to desktop computers.
+
+Different operating systems have distinct methods of accessing emoji. Note that these methods work in most applications, not just WordPress.
+
+## Why remove emoji?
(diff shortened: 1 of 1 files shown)


docs/remove-emoji.md
```diff
@@ -0,0 +1,27 @@
+# Remove Emoji
```
roborourke, line 1: I'm thinking we should label this as "Remove Emoji Image Replacement".

Removing emoji isn't really what we're doing here. It's a tricky one really, I'm not 100% sure on the best wording...

jazzsequence: Sounds good to me. I went back and forth on what to put here and fell back to just how the component/plugin was named, but `Remove Emoji Image Replacement` is a lot more clear.

roborourke: Yeah that's better. I'm still thinking about the overall content, it reads a bit like emoji are being disabled which isn't quite right. Need to focus on the image replacement aspect of it more I think.

jazzsequence: I guess I misunderstood what the module was doing. It's not obvious in the code or inline documentation what the intent is and what's actually happening.

roborourke:
I'm kind of in 2 minds here. In the original issue I think I chose a bad name for the config option. Talking about removing emoji is a bit confusing and drastic sounding as we're really just removing the image replacement functionality. We'd need to be very clear about that in the wording throughout.

Do you think the general explanation of what emoji are and the details about the specific JS files that are no longer loaded might be more than is really needed? I think it just needs to explain that if your website might be run on systems that do not support emoji you can enable image replacements by setting this setting.

In a similar pattern to the search and SEO module docs maybe we could condense this into a small section on the main readme if there isn't enough to warrant a full page.


I might make an issue for renaming that config option first (we can keep back compat) but I think it'll help with making these docs clearer.

jazzsequence:
@roborourke I think renaming the config option is a better first step. As you say, "remove emoji" sounds pretty drastic and misleading, and the description, I think, makes less sense if you go into it with that frame of reference. As such, we should punt this to 6.0 and make changing the option name a dependency of this.

docs/remove-emoji.md
```diff
@@ -0,0 +1,27 @@
+# Remove Emoji Image Replacement
+
+Remove all the emoji scripts that are included with WordPress. By default, this module is active. To allow emojis, toggle the `remove-emoji` module in the Altis config.
```
jazzsequence, line 3: "To allow emojis" should be "To allow emoji image replacement"

roborourke:
Created #260 for the setting name change

joehoyle:
Closing as stale

Issue #204, opened 2020-03-13
WordPress 5.4


rmccue:
Adding a tracking issue for the WordPress 5.4 update. We'll likely want to include this with v4, and backport as preview to v3 and v2 (v1 will be End Of Life at this point I think).

What customisations do we want to apply?

rmccue:
I'm thinking we should disable the [fullscreen block editor by default](https://make.wordpress.org/core/2020/03/03/fullscreen-mode-enabled-by-default-in-the-editor/) change; there's [a snippet here](https://jeanbaptisteaudras.com/en/2020/03/disable-block-editor-default-fullscreen-mode-in-wordpress-5-4/) which should let us do that.

joehoyle:
Given that Altis 4 looks looks some time away, I think we should get this backported to preview releases ASAP.

igmoweb:
Yes, would be nice to have it as soon as possible. There are a few new features in Gutenberg that could benefit Concentra project.

joehoyle:
It looks like the decision for full-screen by default is only for new users, not sure if that changes what we want to do there.

glueckpress:
> It looks like the decision for full-screen by default is only for new users

Nope, had it go fullscreen on my personal blog with a 10+ years old user account right after the update.

mikeselander:
Making this available to 3.0 sooner rather than later would be ace ✨

roborourke:
@mikeselander any specific things that are needed? We're keeping the patch releases up to date, just asking as it's a fair bit of overhead to maintain and test a preview release. Note that WP 5.4 won't be added to 3.0 stable ever, it'd be 3.1.

roborourke:
Adding a note here that I made the following changes for 5.4:

- #219 hide the healthcheck dashboard widget
- #218 remove the default full screen mode and welcome guide behaviour

Everything else seemed to be fine so far.

mikeselander:
> @mikeselander any specific things that are needed? We're keeping the patch releases up to date, just asking as it's a fair bit of overhead to maintain and test a preview release. Note that WP 5.4 won't be added to 3.0 stable ever, it'd be 3.1.

3.1 would be fine, my need was mostly temporary and has now gone away (needed to test a client migration on to Altis locally, but they had 5.4) but it would generally be a good thing to have 5.4 available for 3.0 users since 4.0 is such a large change and imo it should be possible to roll out a major WP update within a few weeks of its release.

PR #218, opened 2020-06-04
Change the WP 5.4 default block editor settings

roborourke:
In WP 5.4 the block editor will by default open in full screen mode with the welcome guide showing.

This update sets both of those defaults to false and allows for persisting settings users have opted in to.


roborourke:
Related to https://github.com/humanmade/altis-cms/issues/204

PR #219, opened 2020-06-04
Hide healthcheck dashboard widget

roborourke:
WP 5.4 introduces a new dashboard widget with the site healthcheck, because we use custom healthchecks this is misleading for Altis so we need to remove it by default.


roborourke:
Related to https://github.com/humanmade/altis-cms/issues/204

Issue #168, opened 2019-09-10
Add apcu-autoloader composer flag


joehoyle:
This could be interesting for performance reasons in production.

----------

Acceptance criteria:

- [ ] Add the `--apcu-autoloader` flag to the `composer install` line in `.build-script` - do not remove any other arguments!
- [ ] Double check everything still works as expected on both local chassis and local server when adding this flag to the `composer create-project altis/skeleton altis-test` command

PR #232, opened 2020-07-01
Add autoloader flag to command

pdewouters:
Fixes https://github.com/humanmade/altis-cms/issues/168

@joehoyle I'm not clear on how to test the second acceptance criteria. Do you mean run the command locally with that parameter? `composer create-project altis/skeleton altis-test --apcu-autoloader`

### DIFF
--- a/.build-script
+++ b/.build-script
@@ -5,7 +5,7 @@
 #====================#
 
 # Install composer dependencies
-composer install --no-dev --optimize-autoloader
+composer install --no-dev --optimize-autoloader --apcu-autoloader
 
 # Edit below this line to add any additional build steps.
 #


joehoyle:
@pdewouters yeah that's what I was thinking, assuming that is a valid param to `create-project` not just `install`.

pdewouters:
Screenshot 2020-07-01 at 16 06 45

Doesn't look like it does

joehoyle:
Ok, in that case I think we just want to confirm everythign works ok if you run `rm -rf vendor ; composer install --apcu-autloader`

pdewouters:
The site loads ok - are there specific tests to run through?

joehoyle:
@pdewouters I think that's all good then!

PR #233, opened 2020-07-03
Remove default WP favicon

rmccue:
> By default, WordPress will send its own logo for favicon requests.
> favicon.ico is blocked by our Cloud servers, but local installs will
> incorrectly show the WordPress logo due to this fallback.
>
> This function replaces the fallback with a zero-byte icon file, which was
> the fallback prior to WordPress 5.4. (Sites can continue to use the site
> icon functionality to set their own icons instead.)

See https://make.wordpress.org/core/2020/02/19/enhancements-to-favicon-handling-in-wordpress-5-4/

We missed this during the 5.4 upgrade; I suspect local-server matches the Cloud behaviour?


joehoyle:
> I suspect local-server matches the Cloud behaviour?

Correct.

PR #237, opened 2020-07-09
Use real GUIDs

### LINKED ISSUE #229: Add real GUIDs
Default GUIDs in WordPress are based off the current site URL and are generally a poor indicator of uniqueness when you factor in migrations and the potential for breakages on search / replace operations etc...

We should switch to generating real GUIDs for posts (WP even has a built in function for this now).

This change won't affect existing content and will only apply to new content.

@rmccue has a plugin here https://github.com/rmccue/realguids

----------------------------

Acceptance criteria:

- [ ] Create a new namespace folder and file under `inc` and include in the `composer.json` autoload files section
- [ ] Copy the filters from the plugin code linked above into a setup function
- [ ] Call the setup function from the CMS module set up function
- [ ] Replace the linked plugin's `generate_uuid()` function with the built in `wp_generate_uuid()`

pdewouters:
fixes #229


inc/real_guids/namespace.php
```diff
@@ -0,0 +1,36 @@
+<?php
+
+namespace Altis\CMS\Real_GUIDs;
+
+function bootstrap() {
```
roborourke, line 5: Missing docblock here:
```suggestion
/**
 * Add filters for setting real GUIDs.
 */
function bootstrap() {
```

inc/real_guids/namespace.php
```diff
@@ -0,0 +1,36 @@
...
+}
+
+/**
+ * Add our UUID to new posts.
+ *
+ * @param array $data Post data to save to the database.
+ * @return array
+ */
+function add_uuid_to_new( array $data ) : array {
+	// Set a default GUID
```
roborourke, line 30: Comments should end with a full stop.
```suggestion
	// Set a default GUID.
```

PR #243, opened 2020-07-24
Override sites list for better detail

rmccue:
The sites list has an unclear visual hierarchy, and highlights less relevant details for most users.

This PR changes the site list to have site names as the primary entry instead of the URL, and de-emphasises less-important information. Additionally, it hides the rarely-used Spam action primarily used for WordPress.com-style open-registration multisites.

Before | After
-- | --
Screenshot_2020-07-24 Sites ‹ Network Admin Altis Demo — Altis(3) | Screenshot_2020-07-24 Sites ‹ Network Admin Altis Demo — Altis(2)

Still needs docs for the additional setting.

### DIFF
--- a/inc/network_ui/namespace.php
+++ b/inc/network_ui/namespace.php
@@ -107,7 +107,7 @@ function render_name_column( $id ) : void {
 	printf(
 		'<strong class="altis-site-name"><a href="%s" class="edit" title="%s">%s</a></strong>',
 		esc_url( get_admin_url( $id ) ),
-		__( 'Open site dashboard', 'altis' ),
+		esc_html__( 'Open site dashboard', 'altis' ),
 		esc_html( $name )
 	);
 	printf(
@@ -120,7 +120,7 @@ function render_name_column( $id ) : void {
 	if ( 'list' !== $mode ) {
 		printf(
 			'<p><em>%s</em></p>',
-			$desc
+			esc_html( $desc )
 		);
 	}
 }
@@ -132,7 +132,7 @@ function render_name_column( $id ) : void {
  * @return void Outputs directly.
  */
 function render_site_states( WP_Site $site ) : void {
-	$site_states = array();
+	$site_states = [];
 	$wp_list_table = _get_list_table( 'WP_MS_Sites_List_Table' );
 
 	if ( is_main_site( $site->id ) ) {
@@ -168,8 +168,11 @@ function render_site_states( WP_Site $site ) : void {
 		echo ' <em>(';
 		foreach ( $site_states as $state ) {
 			++$i;
-			( $i == $state_count ) ? $sep = '' : $sep = ', ';
-			echo "<span class='post-state'>{$state}{$sep}</span>";
+			printf(
+				'<span class="post-state">%s%s</span>',
+				esc_html( $state ),
+				$i === $state_count ? '' : ', '
+			);
 		}
 		echo ')</em>';
 	}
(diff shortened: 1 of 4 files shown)


inc/network_ui/namespace.php
```diff
@@ -0,0 +1,174 @@
...
+	$site_states = apply_filters( 'display_site_states', $site_states, $site );
+
+	if ( ! empty( $site_states ) ) {
+		$state_count = count( $site_states );
+		$i           = 0;
+		echo ' <em>(';
+		foreach ( $site_states as $state ) {
+			++$i;
+			( $i == $state_count ) ? $sep = '' : $sep = ', ';
+			echo "<span class='post-state'>{$state}{$sep}</span>";
```
rmccue, line 170: damn it Wordpress

inc/network_ui/namespace.php
```diff
@@ -0,0 +1,174 @@
...
+
+	$shorturl = untrailingslashit( $blog->domain . $blog->path );
+	if ( $blog->path !== '/' ) {
+		$shorturl = trailingslashit( $shorturl );
+	}
+
+	// Gather information.
+	switch_to_blog( $blog->id );
+	$name = get_option( 'blogname' );
+	// $edit = ...;
```
roborourke, line 102: 🤔

inc/network_ui/namespace.php
```diff
@@ -0,0 +1,174 @@
...
+
+	if ( 'list' !== $mode ) {
+		printf(
+			'<p><em>%s</em></p>',
+			$desc
+		);
+	}
+}
+
+function render_site_states( WP_Site $site ) {
```
roborourke, line 129: Docblock plz

inc/network_ui/namespace.php
```diff
@@ -0,0 +1,174 @@
...
+	add_filter( 'manage_sites_action_links', __NAMESPACE__ . '\\change_row_actions', 0, 2 );
+}
+
+/**
+ * Get configuration for the network UI.
+ *
+ * @return array
+ */
+function get_config() : array {
+	return Altis\get_config()['modules']['cms']['network_ui'];
```
roborourke, line 23: We have some inconsistency across module settings - we started out with mostly hyphenated setting names I think. Would be good to double check and do a sweep to standardise these.

inc/network_ui/namespace.php
rmccue: whups

inc/network_ui/namespace.php
rmccue: whups

inc/network_ui/namespace.php
roborourke: hwhups

inc/network_ui/namespace.php
```diff
@@ -0,0 +1,179 @@
...
+		'backend' => $actions['backend'],
+	];
+	unset( $actions['backend'] );
+	$actions = array_merge( $prefix, $actions );
+
+	$config = get_config();
+	$site = get_site( $id );
+
+	// Hide "Spam", unless the site is marked as spam (i.e. action is "Not Spam")
+	if ( $config['disable-spam'] && ! $site->spam ) {
```
roborourke, line 78: The value of `$config` isn't guaranteed to be an array here, that's just the default.

roborourke (changes requested):
One minor potential bug, otherwise good to go

inc/network_ui/namespace.php
rmccue: That'd fail the return type check from `get_config()` above

PR #248, opened 2020-07-29
Add post cloner plugin

pdewouters:
@roborourke this depends on https://github.com/humanmade/post-cloner/pull/25 - so we'll probably have to bump the version as well?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
     "altis/cms-installer": "^0.4.2",
     "johnbillion/extended-cpts": "^4.4.0",
     "humanmade/clean-html": "^2.0.0",
-    "humanmade/post-cloner": "1.1.1"
+    "humanmade/post-cloner": "~1.1.1"
   },
   "autoload": {
     "files": [
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,8 @@
     "johnpbloch/wordpress": "5.4.2",
     "altis/cms-installer": "^0.4.2",
     "johnbillion/extended-cpts": "^4.4.0",
-    "humanmade/clean-html": "^2.0.0"
+    "humanmade/clean-html": "^2.0.0",
+    "humanmade/post-cloner": "^1.0.0"
   },
   "autoload": {
     "files": [
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
     "altis/cms-installer": "^0.4.2",
     "johnbillion/extended-cpts": "^4.4.0",
     "humanmade/clean-html": "^2.0.0",
-    "humanmade/post-cloner": "^1.0.0"
+    "humanmade/post-cloner": "1.1.1"
   },
   "autoload": {
     "files": [
--- a/docs/post-cloner.md
+++ b/docs/post-cloner.md
@@ -26,46 +26,46 @@ Fires when post meta has been copied to cloned post.
 
 ## Filters
 
-**`post_cloner_cloned_parent: (int) $parent Parent - defaults to original post ID.`**
+**`post_cloner_cloned_parent: (int) $parent - defaults to original post ID`**
 
 Allows overriding the parent that the new post should have.
 
-**`post_cloner_cloned_status: (string) $status Status - defaults to draft.`**
+**`post_cloner_cloned_status: (string) $status - defaults to 'draft'`**
 
 Allows overriding the status that the new post should have.
 
-**`post_cloner_name_append: (string) $suffix String to append to the end of a cloned post's name. - defaults to '-cloned'`**
+**`post_cloner_name_append: (string) $suffix - defaults to '-cloned'`**
 
 Allows overriding the string to append onto the end of post name to prevent collisions.
 
-**`post_cloner_meta_keys_to_remove: (array) $keys_to_remove Metadata keys.`**
+**`post_cloner_meta_keys_to_remove: (array) $keys_to_remove - defaults to ['_edit_lock, '_edit_last']`**
 
 Allows modifying the denylist of keys of metadata that should not be passed through to the new post.
 
-**`post_cloner_meta_patterns_to_remove: (array) $patterns Regex patterns to look for when cleaning meta keys.`**
+**`post_cloner_meta_patterns_to_remove: (array) $patterns - default is empty array`**
 
 Allows cleaning meta keys that match a regular expression.
 
-**`post_cloner_meta_data: (array) $post_meta All post meta to assign to the new post.`**
+**`post_cloner_meta_data: (array) $post_meta - defaults to assoc array with keys 'post_cloned' and 'post_cloned_from'`**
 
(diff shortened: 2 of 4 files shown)


composer.json
```diff
@@ -15,7 +15,8 @@
     "johnpbloch/wordpress": "5.4.2",
     "altis/cms-installer": "^0.4.2",
     "johnbillion/extended-cpts": "^4.4.0",
-    "humanmade/clean-html": "^2.0.0"
+    "humanmade/clean-html": "^2.0.0",
+    "humanmade/post-cloner": "^1.0.0"
```
roborourke, line 19: Let's set this to `~1.0.0` so it just tracks patch versions - I'm starting to update the version pinning for our plugins in this way across the board from now on.

pdewouters: actually it will be `1.1.1`

docs/post-cloner.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Actions
+
+**`post_cloner_after_meta_data_copy: (int) $post_id`**
+
+Fires when post meta has been copied to cloned post.
+
+## Filters
+
+**`post_cloner_cloned_parent: (int) $original_post['ID']`**
```
roborourke, line 29: Seems odd to show the variable that is actually passed to this filter - when using the filter you would name this parameter `$post_id` I think - it doesn't need to match the way it's called directly.

docs/post-cloner.md
```diff
@@ -0,0 +1,71 @@
...
+
+Allows overriding the string to append onto the end of post name to prevent collisions
+
+**`post_cloner_meta_keys_to_remove: (array) $keys_to_remove`**
+
+Allows modifying the denylist of keys of metadata that should not be passed through to the new post.
+
+**`post_cloner_meta_patterns_to_remove: (array) []`**
+
+Allows overriding the regex patterns to look for when cleaning meta keys
```
roborourke, line 47: Can we explain this one a little more clearly, perhaps with an example value? I'm not sure what this means exactly, but guessing it's related to the filter above.

docs/post-cloner.md
```diff
@@ -0,0 +1,71 @@
...
+
+**`post_cloner_cloned_status: (string) 'draft'`**
+
+Allows overriding the status that the new post should have
+
+**`post_cloner_name_append: (string) '-cloned'`**
+
+Allows overriding the string to append onto the end of post name to prevent collisions
+
+**`post_cloner_meta_keys_to_remove: (array) $keys_to_remove`**
```
roborourke, line 41: Some of these show the default values but in some cases it's a variable name. Can we take a consistent approach eg. a variable name followed by its default value for each filter?

Also can you add an issue on the upstream repo - I think some of these filters would benefit from passing the `$post_id` as well as a second parameter.

docs/post-cloner.md
```diff
@@ -0,0 +1,71 @@
...
+
+Allows overriding the allowlist of post types that are eligible for cloning.
+
+**`post_cloner_clonable_statuses: (array) [ 'publish', 'draft', 'pending' ]`**
+
+Allows overriding the allowlist of post statuses that are eligible for cloning.
+
+**`post_cloner_permission_level: (string) 'publish_posts'`**
+
+Allows overriding of minimum capability that a user must have to clone a post.
```
roborourke, line 71:
```suggestion
Allows overriding the minimum capability that a user must have to clone a post.
```

It's a small thing but try to avoid the kind of writing we might do in code comments in the docs, this one read a little strangely.

docs/post-cloner.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Filters
+
+**`post_cloner_cloned_parent: (int) $original_post['ID']`**
+
+Allows overriding the parent that the new post should have.
+
+**`post_cloner_cloned_status: (string) 'draft'`**
+
+Allows overriding the status that the new post should have
```
roborourke, line 35: Make sure we're consistent in the use of full stops at the end of sentences, some of these do and some don't.

composer.json
```diff
@@ -15,7 +15,8 @@
     "johnpbloch/wordpress": "5.4.2",
     "altis/cms-installer": "^0.4.2",
     "johnbillion/extended-cpts": "^4.4.0",
-    "humanmade/clean-html": "^2.0.0"
+    "humanmade/clean-html": "^2.0.0",
+    "humanmade/post-cloner": "1.1.1"
```
roborourke, line 19:
```suggestion
    "humanmade/post-cloner": "~1.1.1"
```

Let's track the patch release version

docs/post-cloner.md
pdewouters: I'm not sure either, I suppose it's if you want to skip copying all meta data with a key like `foo_`

docs/post-cloner.md
```diff
@@ -0,0 +1,71 @@
...
+
+**`post_cloner_cloned_parent: (int) $parent Parent - defaults to original post ID.`**
+
+Allows overriding the parent that the new post should have.
+
+**`post_cloner_cloned_status: (string) $status Status - defaults to draft.`**
+
+Allows overriding the status that the new post should have.
+
+**`post_cloner_name_append: (string) $suffix String to append to the end of a cloned post's name. - defaults to '-cloned'`**
```
roborourke, line 37: I think these get a bit lengthy - rather than have a description of the parameter inline I was thinking just the parameter name and the default value if there is one. The explanation of the first parameter is effectively covered in the filter description below.

docs/post-cloner.md
roborourke: Could you look into how it's used in the post-cloner code so we can be sure?

docs/post-cloner.md
```diff
@@ -0,0 +1,72 @@
...
+
+**`post_cloner_cloned_status: (string) $status - defaults to 'draft'`**
+
+Allows overriding the status that the new post should have.
+
+**`post_cloner_name_append: (string) $suffix - defaults to '-cloned'`**
+
+Allows overriding the string to append onto the end of post name to prevent collisions.
+
+**`post_cloner_meta_keys_to_remove: (array) $keys_to_remove - defaults to ['_edit_lock, '_edit_last']`**
```
roborourke:
```suggestion
**`post_cloner_meta_keys_to_remove: (array) $keys_to_remove - defaults to [ '_edit_lock, '_edit_last' ]`**
```

docs/post-cloner.md
```diff
@@ -0,0 +1,72 @@
...
+
+Allows modifying the denylist of keys of metadata that should not be passed through to the new post.
+
+**`post_cloner_meta_patterns_to_remove: (array) $patterns - default is empty array`**
```
roborourke:
```suggestion
**`post_cloner_meta_patterns_to_remove: (array) $patterns - defaults to []`**
```

roborourke:
Thanks @pdewouters, merging this now.

Issue #238, opened 2020-07-15
Activate email doesn't use wp_mail_from


joehoyle:
It seems as though the email send for account activate "Activate $name" does not use the value from `apply_filters( 'wp_mail_from' )`, this means our overrides in altis-cloud don't work. It uses the value of `get_site_option( ‘admin_email’ );` and falls back to `$admin_email = 'support@' . $_SERVER['SERVER_NAME'];` which I think is not a good default for us. Atleast we could fallback to the wp_mail_from value.

props @stuartshields for digging into this.

stuartshields:
Link to the line of code causing the issue:
https://github.com/WordPress/WordPress/blob/master/wp-includes/ms-functions.php#L952-L956

roborourke:
Fixed in #249

PR #249, opened 2020-08-03
Replace WP blog signup notification

joeleenk:
Removes the default `wpmu_signup_blog_notification` and replaces it with an Altis one, so that we can control the email address the notices are sent from. 

See https://github.com/humanmade/altis-cms/issues/238

### DIFF
--- a/inc/signup_notification/namespace.php
+++ b/inc/signup_notification/namespace.php
@@ -1,0 +1,155 @@
+<?php
+/**
+ * Altis CMS Sign-up Blog Notification.
+ *
+ * @package altis/cms
+ */
+
+namespace Altis\CMS\Signup_Notification;
+
+/**
+ * Boostrap setup to remove updates from the admin.
+ */
+function bootstrap() {
+	remove_action( 'after_signup_site', 'wpmu_signup_blog_notification', 10, 7 );
+	add_action( 'after_signup_site', __NAMESPACE__ . '\\altis_signup_blog_notification', 10, 7 );
+}
+
+/**
+ * Send a confirmation request email to a user when they sign up for a new site. The new site will not become active
+ * until the confirmation link is clicked.
+ * 
+ * NOTE: This is a replacement function for the core wpmu_signup_blog_notification function:
+ * https://github.com/WordPress/WordPress/blob/master/wp-includes/ms-functions.php#L900-L1036
+ *
+ * This is the notification function used when site registration
+ * is enabled.
+ *
+ * Filter {@see 'wpmu_signup_blog_notification'} to bypass this function or
+ * replace it with your own notification behavior.
+ *
+ * Filter {@see 'wpmu_signup_blog_notification_email'} and
+ * {@see 'wpmu_signup_blog_notification_subject'} to change the content
+ * and subject line of the email sent to newly registered users.
+ *
+ * @since MU (3.0.0)
+ *
+ * @param string $domain     The new blog domain.
+ * @param string $path       The new blog path.
+ * @param string $title      The site title.
+ * @param string $user_login The user's login name.
+ * @param string $user_email The user's email address.
+ * @param string $key        The activation key created in wpmu_signup_blog()
+ * @param array  $meta       Optional. Signup meta data. By default, contains the requested privacy setting and lang_id.
+ * @return bool
+ */
+function altis_signup_blog_notification( $domain, $path, $title, $user_login, $user_email, $key, $meta = array() ) {
+	/**
+	 * Filters whether to bypass the new site email notification.
+	 *
+	 * @since MU (3.0.0)
+	 *
+	 * @param string|bool $domain     Site domain.
+	 * @param string      $path       Site path.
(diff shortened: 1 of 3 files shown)


inc/signup_notification/namespace.php
```diff
@@ -0,0 +1,155 @@
...
+ * @package altis/cms
+ */
+
+namespace Altis\CMS\Signup_Notification;
+
+/**
+ * Boostrap setup to remove updates from the admin.
+ */
+function bootstrap() {
+	remove_action( 'after_signup_site', 'wpmu_signup_blog_notification', 10, 7 );
```
roborourke, line 14: Is this working ok? I thought this code might run before the default action has been added in WP core.

inc/signup_notification/namespace.php
```diff
@@ -0,0 +1,155 @@
...
+		$activate_url = "http://{$domain}{$path}wp-activate.php?key=$key"; // @todo Use *_url() API.
+	}
+
+	$activate_url = esc_url( $activate_url );
+
+	$admin_email = get_site_option( 'admin_email' );
+
+	$admin_email = apply_filters( 'wp_mail_from', $admin_email );
+
+	$from_name       = ( '' !== get_site_option( 'site_name' ) ) ? esc_html( get_site_option( 'site_name' ) ) : 'WordPress';
```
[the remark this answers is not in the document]
joeleenk: 👍

inc/signup_notification/namespace.php
```diff
@@ -0,0 +1,155 @@
...
+	 *
+	 * @param string|bool $domain     Site domain.
+	 * @param string      $path       Site path.
+	 * @param string      $title      Site title.
+	 * @param string      $user_login User login name.
+	 * @param string      $user_email User email address.
+	 * @param string      $key        Activation key created in wpmu_signup_blog().
+	 * @param array       $meta       Signup meta data. By default, contains the requested privacy setting and lang_id.
+	 */
+	if ( ! apply_filters( 'altis_signup_blog_notification', $domain, $path, $title, $user_login, $user_email, $key, $meta ) ) {
```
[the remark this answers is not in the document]
joeleenk: I agree, I wasn't sure on this one, but I was thinking we might still need the original filter name. Thanks for helping clarify it.

inc/signup_notification/namespace.php
joeleenk: I changed the remove function priority to make sure it runs after it's added. [0eda048](https://github.com/humanmade/altis-cms/pull/249/commits/0eda0489b3aa3cea29a9923ada782b58b0dced80)
For reference, it's [added here](https://github.com/WordPress/WordPress/blob/6496c6174148436cad3211a6392f34953657af89/wp-includes/ms-default-filters.php#L42).

roborourke: Awesome, thanks @joeleenk

inc/signup_notification/namespace.php
```diff
@@ -0,0 +1,162 @@
...
+ */
+
+namespace Altis\CMS\Signup_Notification;
+
+/**
+ * Boostrap setup to remove updates from the admin.
+ */
+function bootstrap() {
+	add_action( 'after_signup_site', __NAMESPACE__ . '\\altis_signup_blog_notification', 10, 7 );
+	add_action( 'after_signup_site', __NAMESPACE__ . '\\remove_wpmu_signup_blog_notification', 11 );
```
roborourke, line 15: Hmm, this will fire _after_ anything hooked on priority 10 so it should be on priority 9 or lower.
```suggestion
	add_action( 'after_signup_site', __NAMESPACE__ . '\\remove_wpmu_signup_blog_notification', 9 );
```

roborourke:
Good work 😄

inc/signup_notification/namespace.php
joeleenk: Don't we want it to fire after the `add_action`? You can't remove it if it's not there yet.

roborourke: The default action callback in core is not added during the `after_signup_site` hook but before that hook is ever run in `ms-default-filters.php`.

When an action hook is being processed it will execute all attached callbacks in the order of a) priority and b) the order they were added. This means that using the priority 11 would have resulted in the following chain of events:

1. `ms-default-filters.php` loaded, `after_signup_site` default notification hook is added
2. `after_signup_site` action is triggered
3. Default callback is run on priority 10
4. Callback to remove the default notification is run on priority 11

In this scenario the default notification is removed too late as it would have already been run. By changing the priority to 9 the callback to remove the default notification is run before it reaches the callbacks attached at priority 10.

I think the confusion was caused by using the same action hook to do the removal - the default notification hook is added when `ms-default-filters.php` is included just before the `muplugins_loaded` hook is triggered, so you could have used that or any hook after that eg. such as `init`.

joeleenk: Thanks for walking me through this! I was having a hard time working my head around the order, this helps. 😄

PR #250, opened 2020-08-05
Delete signups object cache before loading the signups page

### LINKED ISSUE #84: When users are added from a site, they don't show up in Sign Ups
The Sign Ups admin page seems to use the object cache, and when a user is added directly to a site (without skipping confirmation) they don't show up in the network admin Sign Ups list until the object cache is flushed.

joeleenk:
The WP User Signups plugin uses the object cache, so when a user is added directly to a site (without skipping confirmation) they don't show up in the network admin Sign Ups list until the object cache is flushed. This deletes the cache contents for the signups key/group before fetching users so that the plugin is forced to load the new ones. 

Fixes #84


inc/namespace.php
```diff
@@ -114,6 +114,9 @@ function bootstrap() {
 	 */
 	add_filter( 'script_loader_src', __NAMESPACE__ . '\\real_url_path', -10, 2 );
 	add_filter( 'style_loader_src', __NAMESPACE__ . '\\real_url_path', -10, 2 );
+
+	// Delete signups object cache before we load the signups page.
+	add_action( 'pre_get_signups', __NAMESPACE__ . '\\altis_pre_get_signups' );
```
roborourke, line 119: I think Altis is implicit here so we can change the function prefix.

inc/namespace.php
```diff
@@ -390,3 +393,26 @@ function real_url_path( string $url, string $handle ) : string {
 
 	return $url;
 }
+
+/**
+ * Delete signups object cache before we load the signups page.
+ *
+ * @param object $signup_query The current instance of the WP_Signup_Query object.
+ */
+function altis_pre_get_signups( $signup_query ) {
```
roborourke, line 402: We can add type hinting here if we know what the object type is eg:
```suggestion
function altis_pre_get_signups( WP_Signup_Query $signup_query ) {
```
Will need to `use` this class as well though.

PR #261, opened 2020-10-09
Add set -e to build script

### LINKED ISSUE #262: The build script can still pass even if a step fails
Steps to reproduce:
1. Add a failing step to the buiid script
2. Build will complete in Vantage

The build should fail to avoid deploying a broken application.

Acceptance criteria:

- [ ] Ensure default build script return a non zero exit code if a build step fails

roborourke:
Deploys can still try and go ahead even if a build step fails. This will make sure that doesn't happen.

Fixes #262

### DIFF
--- a/.build-script
+++ b/.build-script
@@ -4,6 +4,9 @@
 # Altis Build Script #
 #====================#
 
+# Ensure build script fails if any steps return a non-zero exit code.
+set -e
+
 # Install composer dependencies
 composer install --no-dev --optimize-autoloader --apcu-autoloader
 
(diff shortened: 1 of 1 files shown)


.build-script
```diff
@@ -4,6 +4,9 @@
 # Altis Build Script #
 #====================#
 
+# Ensure build script fails if any steps return a non-zero exit code.
+set -e
```
rmccue, line 8: Let's build this into the shebang line instead. Can do `#!/bin/bash -e`

roborourke: Done

PR #282, opened 2020-11-03
set AUTOMATIC_UPDATER_DISABLED to true to disable auto updates

nathanielks:
WordPress' automatic updater expects this constant to return true if auto updates are to be disabled:

https://github.com/WordPress/WordPress/blob/9da0a44b3e70117089257777cc38a3305a6c6c77/wp-admin/includes/class-wp-automatic-updater.php#L41

### DIFF
--- a/inc/remove_updates/namespace.php
+++ b/inc/remove_updates/namespace.php
@@ -48,7 +48,7 @@ function remove_update_nag() {
  * Remove all automatic updates.
  */
 function remove_auto_updates() {
-	define( 'AUTOMATIC_UPDATER_DISABLED', false );
+	define( 'AUTOMATIC_UPDATER_DISABLED', true );
 }
 
 /**


nathanielks:
We received a bug report from a developer noting they've been getting emails every 12 hours, notifying them they've been updated to WordPress 5.4.4. This is because the `wp_version_check` cron event runs `twicedaily`.

rmccue:
We already disable file modifications, which is an earlier check there: https://github.com/WordPress/WordPress/blob/9da0a44b3e70117089257777cc38a3305a6c6c77/wp-admin/includes/class-wp-automatic-updater.php#L32

Could be that they've misconfigured it to override that constant?

rmccue:
Independently of that though, this _does_ seem like a bug in the code as written.

nathanielks:
I looked into `wp_is_file_mod_allowed()` and it doesn't appear we're disabling file mods. I tested this on two of our active environments:

```
$ wp shell
wp> ! wp_is_file_mod_allowed( 'automatic_updater' )
=> bool(false)
wp> $disabled = defined( 'AUTOMATIC_UPDATER_DISABLED' ) && AUTOMATIC_UPDATER_DISABLED;
=> bool(false)
wp> DISALLOW_FILE_MODS
Warning: Use of undefined constant DISALLOW_FILE_MODS - assumed 'DISALLOW_FILE_MODS' (this will throw an Error in a future version of PHP) in phar:///usr/local/bin/wp/vendor/wp-cli/shell-command/src/WP_CLI/Shell/REPL.php(46) : eval()'d code on line 1
=> string(18) "DISALLOW_FILE_MODS"
```

Further, I searched across the codebase and there's no Altis references to `DISALLOW_FILE_MODS` or the `file_mod_allowed` filter. It doesn't appear that we have file mods disabled, at present.

roborourke:
Yeah this is decidedly a bug as written! I'm going to approve this change and we should look at maybe setting the `DISALLOW_FILE_MODS` constant. I thought it would determine this from the file system permissions though.

roborourke (approved):

PR #283, opened 2020-11-05
Fix: login-logo URL can be incorrect for all the sub-sites in a network

### LINKED ISSUE #259: Login logo is 404 for sub-sites
Steps to reproduce:
1. Open a login page for any sub-sites
2. Login Logo will not be visible and the logo URL will be 404 in the developer console.

This is probably because the login logo URL that is generated is dynamically generated based on the site URL of the sub-sites 

https://github.com/humanmade/altis-cms/blob/4271a71d0e715a4977950e715f264393203e114b/inc/namespace.php#L185

Acceptance criteria:

- [ ]  Login logo is visible on all sub-sites

Nikschavan:
Fix #259

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -182,7 +182,7 @@ function add_login_logo() {
 	?>
 	<style>
 		.login h1 a {
-			background-image: url('<?php echo esc_url( site_url( $logo ) ); ?>');
+			background-image: url('<?php echo esc_url( get_site_url( 1, $logo ) ); ?>');
 			background-size: contain;
 			width: auto;
 		}
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -182,7 +182,7 @@ function add_login_logo() {
 	?>
 	<style>
 		.login h1 a {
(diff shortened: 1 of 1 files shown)


Nikschavan:
Wondering if this will be backported to v4 as well?

inc/namespace.php
```diff
@@ -182,7 +182,7 @@ function add_login_logo() {
 	?>
 	<style>
 		.login h1 a {
-			background-image: url('<?php echo esc_url( site_url( $logo ) ); ?>');
+			background-image: url('<?php echo esc_url( get_site_url( 1, $logo ) ); ?>');
```
roborourke, line 185: Can this use `get_main_site_id( get_main_network_id() )` instead? I think 1 is generally correct but in some really bizarre circumstances it's not always the case

roborourke:
```suggestion
			background-image: url('<?php echo esc_url( get_site_url( get_main_site_id( get_main_network_id() ), $logo ) ); ?>');
```

roborourke (changes requested):
Minor update, thanks for doing this!

roborourke (approved):

PR #297, opened 2020-12-15
Don't fatal on null value in real_url_path()

### LINKED ISSUE #296: Admin inaccessible with real_url_path error after Altis 5 upgrade
After upgrading to Altis 5, the admin of a site is failing to load with the following error:

```
Uncaught Error: Argument 1 passed to Altis\CMS\real_url_path() must be of the type string, null given, called in /usr/src/app/wordpress/wp-includes/class-wp-hook.php on line 287 in /usr/src/app/vendor/altis/cms/inc/namespace.php on line 352
Altis\CMS\real_url_path()
wp-includes/class-wp-hook.php:287
apply_filters()
wp-includes/plugin.php:206
apply_filters()
wp-includes/class.wp-scripts.php:328
do_item()
wp-includes/class.wp-dependencies.php:116
do_items()
wp-includes/class.wp-scripts.php:643
do_head_items()
wp-includes/script-loader.php:1841
print_head_scripts()
wp-includes/class-wp-hook.php:287
apply_filters()
wp-includes/class-wp-hook.php:311
do_action()
wp-includes/plugin.php:478
do_action()
wp-admin/admin-header.php:130
require_once()
wp-admin/index.php:111
```

Any page within wp-admin whitescreens with that visible trace (in a dev environment). Frontend of the site appears to function properly.

roborourke:
In some cases `$url` might be `null` when passed to `real_url_path()`. Ideally WP should warn about this so any error output should be handled upstream.

Fixes #296

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -345,13 +345,18 @@ function set_comments_per_page( $value ) : int {
 /**
  * Ensure URLs do not contain any relative paths.
  *
- * @param string $url The dependency URL.
+ * @param string|null $url The dependency URL.
  * @param string $handle The dependency handle.
- * @return string
+ * @return string|null
  */
-function real_url_path( string $url, string $handle ) : string {
+function real_url_path( ?string $url, string $handle ) : ?string {
 	global $wp_scripts, $wp_styles;
-
+	
+	// Avoid odd behaviour if null or empty value is passed.
+	if ( empty( $url ) ) {
+		return $url;
+	}
+ 
 	// Skip if there are no /./ or /../ patterns.
 	if ( strpos( $url, '/.' ) === false ) {
 		return $url;


inc/namespace.php
```diff
@@ -345,13 +345,18 @@ function set_comments_per_page( $value ) : int {
...
+ * @return string|null
  */
-function real_url_path( string $url, string $handle ) : string {
+function real_url_path( ?string $url, string $handle ) : ?string {
 	global $wp_scripts, $wp_styles;
-
+	
+	// Avoid odd behaviour if null or empty value is passed.
+	if ( empty( $url ) ) {
+		return $url;
```
kadamwhite, line 357: Would it make sense to log or trigger an error here with the $handle, to aid teams debugging this?

roborourke: I did think about it but it seems like this is just a convenient place to log the error rather the right place if that makes sense. WordPress itself should warn about invalid URLs provided for assets which we should fix upstream ideally.

kadamwhite (approved):

roborourke:
If it's happening locally you could use those filters to log the problem asset handles and dig into it further from there.

PR #300, opened 2021-01-08
Add packages to install-overrides

rmccue:
Adds WP packages which need their installation path overridden to the config.

See https://github.com/humanmade/altis-core/pull/120.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,6 +36,12 @@
     ]
   },
   "extra": {
-    "altis": {}
+    "altis": {
+      "install-overrides": [
+        "humanmade/post-cloner",
+        "roots/wp-password-bcrypt",
+        "stuttter/wp-user-signups"
+      ]
+    }
   }
 }


composer.json
```diff
@@ -36,6 +36,12 @@
     ]
   },
   "extra": {
-    "altis": {}
+    "altis": {
+      "install-overrides": [
+        "humanmade/post-cloner",
+        "roots/wp-password-bcrypt",
```
roborourke, line 42: This one isn't a WordPress package type so doesn't need special handling, still won't do any harm though.

roborourke (approved):

composer.json
rmccue: My bad; tried to cross-ref against altis/core, but accidentally picked up another one by the looks of it.

roborourke:
Gonna close out in favour of some automation across all modules for adding these.

PR #301, opened 2021-01-11
Add check to redirect_canonical for equivalent query strings

### LINKED ISSUE #302: Endless redirects occurring with different query string formats
Steps to reproduce:
1. Enable Batcache
2. Set a static page for the home page
3. Browse to `your-site.altis.dev/?foo=bar[]&baz=`
4. Quickly browse to that URL again, repeat until redirect is cached by Batcache
5. Endless redirect occurs to reformatted version of query string

When the query string is effectively the same and will be processed to the same cache key in the case of Batcache a redirect should not occur, or not be cached.

Acceptance criteria:

- [ ] Prevent redirection for any equivalent URLs

roborourke:
This prevents unnecessary redirects where the query strings and base URLs are the same but formatted differently. This will prevent a bug with Batcache caching redirects to another URL with the same derived cache key and endlessly redirecting.

It is generally preferable to avoid unneeded redirects anyway so this is patching the WP core behaviour. There is a trac ticket describing the issue here https://core.trac.wordpress.org/ticket/41712

Fixes #302

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -119,6 +119,9 @@ function bootstrap() {
 
 	// Delete signups object cache before we load the signups page.
 	add_action( 'after_signup_user', __NAMESPACE__ . '\\clear_signups_cache' );
+
+	// Fix redirect canonical redirecting on equivalent query strings.
+	add_filter( 'redirect_canonical', __NAMESPACE__ . '\\maybe_redirect', 11, 2 );
 }
 
 /**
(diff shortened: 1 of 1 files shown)


jazzsequence (approved):

inc/namespace.php
```diff
@@ -410,3 +413,46 @@ function real_url_path( ?string $url, string $handle ) : ?string {
...
+ * a redirect that will generate the same Batcache key. Once the redirect is cached then
+ * Batcache will start to redirect endlessly. This filter prevents that behaviour.
+ *
+ * @param string $redirect_url The URL being redirected to.
+ * @param string $requested_url The original URL requested.
+ * @return string The filtered redirect target URL.
+ */
+function maybe_redirect( $redirect_url, $requested_url ) {
+	$redirect_query_string = parse_url( $redirect_url, PHP_URL_QUERY );
+	$requested_query_string = parse_url( $redirect_url, PHP_URL_QUERY );
```
joehoyle, line 431: @roborourke I'm investigating a similar bug to this, I'm not sure it's exactly the same. However, I noticed this code looks wrong? You are calling `parse_url()` on the same variable each time, when I assume it's meant to be on `redirect_url` and `requested_url`?

roborourke: Yeah that does look wrong, thanks for flagging!

PR #320, opened 2021-01-27
Fix remove emoji DNS prefetch warning

### LINKED ISSUE #272: PHP warning in disable_emojis_remove_dns_prefetch()
```
[29-Oct-2020 12:04:00 UTC] PHP Warning:  strpos() expects parameter 1 to be string, array given in /usr/src/app/vendor/altis/cms/inc/namespace.php on line 165
```

The filter `wp_resource_hints` can return a nested array as well which is not accounted for in the function `disable_emojis_remove_dns_prefetch()`

For eg when checking how WordPress reads the data from this filter, [it is checked if it is an array process it differently in that case](https://core.trac.wordpress.org/browser/tags/5.5.1/src/wp-includes/general-template.php#L3339).

Acceptance criteria:

- [ ] ...

roborourke:
Fixes #272

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -191,6 +191,13 @@ function disable_emojis_remove_dns_prefetch( array $urls, string $relation_type 
 	// Strip out any URLs referencing the WordPress.org emoji location.
 	$emoji_svg_url_bit = 'https://s.w.org/images/core/emoji/';
 	foreach ( $urls as $key => $url ) {
+		if ( is_array( $url ) ) {
+			if ( isset( $url['href'] ) ) {
+				$url = $url['href'];
+			} else {
+				continue;
+			}
+		}
 		if ( strpos( $url, $emoji_svg_url_bit ) !== false ) {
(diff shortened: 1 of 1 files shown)


inc/namespace.php
```diff
@@ -191,6 +191,13 @@ function disable_emojis_remove_dns_prefetch( array $urls, string $relation_type
 	// Strip out any URLs referencing the WordPress.org emoji location.
 	$emoji_svg_url_bit = 'https://s.w.org/images/core/emoji/';
 	foreach ( $urls as $key => $url ) {
+		if ( is_array( $url ) ) {
+			if ( isset( $url['href'] ) ) {
+				$url = $url['href'];
+			} else {
+				continue;
+			}
```
rmccue, line 199: I'd probably flip this to focus on the happy path and reduce nesting.
```suggestion
			if ( ! isset( $url['href'] ) ) {
				continue;
			}

			$url = $url['href'];
```

rmccue (approved):

PR #324, opened 2021-01-28
Fix user signups action link redirect loop

roborourke:
The action links in the user signups UI relied on the value `wp_get_referer()` during processing. As this is missing in cloud environments the logic would result in an endless redirect loop.

This passes the referer via the URL query string and allows it to function normally.

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -92,6 +92,9 @@ function bootstrap() {
 	// Fix network admin site actions.
 	add_filter( 'network_admin_url', __NAMESPACE__ . '\\fix_network_action_confirmation' );
 
+	// Fix user signups admin links.
+	add_filter( 'wp_signups_admin_url', __NAMESPACE__ . '\\fix_user_signups_action_links', 10, 3 );
+
 	// Don't show the welcome panel.
 	add_filter( 'get_user_metadata', __NAMESPACE__ . '\\hide_welcome_panel', 10, 3 );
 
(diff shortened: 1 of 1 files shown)


inc/namespace.php
```diff
@@ -145,6 +148,26 @@ function fix_network_action_confirmation( string $url ) : string {
...
+ *
+ * @param string $url The full URL.
+ * @param string $admin_url The base admin URL.
+ * @param array $args The page query arguments.
+ * @return string
+ */
+function fix_user_signups_action_links( string $url, string $admin_url, array $args ) : string {
+	if ( isset( $args['action'] ) || isset( $args['bulk_action'] ) || isset( $args['bulk_action2'] ) ) {
+		$url = add_query_arg( [
+			'_wp_http_referer' => urlencode( wp_unslash( $_SERVER['REQUEST_URI'] ) ),
```
jazzsequence, line 165: Since this is the same fix as https://github.com/humanmade/altis-cms/pull/316/files, could we kill two birds with one stone and use a single function that handles both URLs? e.g. if `$args` is passed, then we pull the action/bulk action out of that array, and otherwise we do the `wp_parse_url`?

roborourke: I considered it but there’s a subtle difference in that for the other one we’re only interested in the confirm action

roborourke: Also it throw a warning if we accepted additional arguments on the network_admin_url filter

jazzsequence: I wasn't thinking we'd accept additional arguments, more that those may or may not be there but they wouldn't be accepted. I saw the difference in the two functions, just trying to find ways to keep it DRY. Maybe the 

```php
		$url = add_query_arg( [
			'_wp_http_referer' => urlencode( wp_unslash( $_SERVER['REQUEST_URI'] ) ),
		], $url );
```
part could be broken out, although at that point I'm not sure it's worth it. Anyway, just a thought. I don't see any other reason to not push this through.

jazzsequence (approved):

PR #333, opened 2021-02-02
Better sanitization and validation for subdomains, subdirectories and new site URLs

### LINKED ISSUE #303: Create Site URL field allows invalid characters
Steps to reproduce:
1. Go to the create a site page and input "Some Test Name" in the URL field
2. Create the site

Note that the URL is `https://demo.altis.cloud/Some Test Name/`. Attempting to view the site results in a 404.

Sites should not be able to be created with invalid URL components, either by validating the user's input or by sanitising it for them.

Acceptance criteria:

- [x] Custom domains and subdomains with double hyphens work (eg. punycode domains) like xn--rbfka.com
- [ ] Sub directories can include dots, slashes, tilde, hyphens and underscores, and unicode characters are url encoded
- [x] Sub sub domains can be added eg. one.two
- [x] Query strings and hash locations are stripped eg. `?foo=bar` and `#foo`

**Dev Notes**

You'll need to validate based on the type of input selected so:

- Sub folders, sanitise based on valid URL paths, stripping any query string or hash location
- Sub domains sanitise based on valid host name characters
- Custom domains, same as for sub domains but you can further validate that it's a valid hostname eg. it must have a TLD

You could probably construct a full URL and use `sanitize_url()` instead then extract the user input part of the result.

Additionally you could go a bit further and enhance the UX by adding the [`pattern` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/pattern) and changing the value using JS depending on the type of input selected although it might make it difficult to add URLs in other languages like japanese. I'm not sure the extent to which WP supports that but there's some interesting reading here https://www.w3.org/International/articles/idn-and-iri/

jazzsequence:
`sanitize_text_field` does a lot of important work that we want to use for urls. However, what it _doesn't_ do, is disallow text strings from having spaces, which is also important for URLs.

Additionally, there is no handling for IDNAs or other variations that should be allowed. This PR:

* adds URL sanitization that covers additional steps that weren't being handled previously when new sites were created (this code was based off [the old HappyTables code](https://github.com/humanmade/happytables/blob/master/content/plugins-mu/ht-domains/ht-domains.php#L88-L144) for the same)
* pulls in the `Requests_IDNAEncoder` class to properly encode IDNAs
* adds URL encoding for subdirectory sites, so those respect internationalized characters as well (e.g. `domain.com/納豆`)

fixes #303

### DIFF
--- a/inc/add_site_ui/namespace.php
+++ b/inc/add_site_ui/namespace.php
@@ -213,6 +213,22 @@ function sanitize_domain_segment( string $segment ) : string {
 }
 
 /**
+ * Encode a URL string with the IDNA encoder
+ *
+ * @see https://developer.wordpress.org/reference/classes/requests_idnaencoder/
+ * @see https://tools.ietf.org/html/rfc3490
+ *
+ * @uses https://developer.wordpress.org/reference/classes/requests_idnaencoder/encode/
+ *
+ * @param string $url A URL domain string to encode.
+ *
+ * @return string The encoded string.
+ */
+function idna_encode( string $url ) : string {
+	return Requests_IDNAEncoder::encode( $url );
+}
+
+/**
  * Validate a domain name.
  *
  * Checks that the domain name has 2+ valid DNS segments.
--- a/inc/add_site_ui/namespace.php
+++ b/inc/add_site_ui/namespace.php
@@ -441,7 +441,7 @@ function handle_subdirectory( string $value ) : ?array {
 	$network_host = $network_url['host'];
 
 	// Break out URL query parameters and hashes.
-	$path = trim( preg_split( "/(&+|\?+|#+)/", $value )[0], '/' );
+	$path = trim( preg_split( '/(&+|\?+|#+)/', $value )[0], '/' );
 
 	return [
 		'domain' => $network_host,
--- a/inc/add_site_ui/namespace.php
+++ b/inc/add_site_ui/namespace.php
@@ -305,7 +305,7 @@ function add_site_form_handler() {
 		$site_type = 'site-subdomain';
 	}
 
-	$value = sanitize_text_field( wp_unslash( $_POST['url'] ) ?? '' );
+	$value = sanitize_text_field( wp_unslash( $_POST['url'] ?? '' ) );
 	$title = sanitize_text_field( wp_unslash( $_POST['title'] ?? '' ) );
 
 	if ( empty( $value ) || empty( $title ) ) {
--- a/inc/add_site_ui/namespace.php
+++ b/inc/add_site_ui/namespace.php
@@ -203,7 +203,6 @@ function validate_domain_segment( string $segment ) : bool {
  */
 function sanitize_domain_segment( string $segment ) : string {
 	$segment = wp_kses_no_null( $segment );
-	$segment = remove_accents( $segment );
 
 	if ( strlen( $segment ) < 1 ) {
 		return '';
--- a/inc/add_site_ui/namespace.php
+++ b/inc/add_site_ui/namespace.php
@@ -202,7 +202,6 @@ function validate_domain_segment( string $segment ) : bool {
 function sanitize_domain_segment( string $segment ) : string {
 	$segment = wp_kses_no_null( $segment );
 	$segment = remove_accents( $segment );
-	$segment = sanitize_title_with_dashes( $segment, null, 'save' );
 
 	if ( strlen( $segment ) < 1 ) {
 		return '';
--- a/inc/add_site_ui/namespace.php
+++ b/inc/add_site_ui/namespace.php
@@ -278,7 +278,7 @@ function add_site_form_handler() {
 		$site_type = 'site-subdomain';
 	}
 
-	$value = sanitize_text_field( wp_unslash( $_POST['url'] ?? '' ) );
+	$value = sanitize_title( sanitize_text_field( wp_unslash( $_POST['url'] ?? '' ) ) );
 	$title = sanitize_text_field( wp_unslash( $_POST['title'] ?? '' ) );
 
 	if ( empty( $value ) || empty( $title ) ) {
--- a/inc/add_site_ui/namespace.php
+++ b/inc/add_site_ui/namespace.php
@@ -257,9 +257,6 @@ function validate_path( $path ) {
 		return true;
 	}
 
-	// Break out URL query parameters and hashes.
-	$path = preg_split( "/(&+|\?+|#+)/", $path )[0];
-
 	$illegal_names = get_site_option( 'illegal_names' );
 	if ( empty( $illegal_names ) ) {
 		$illegal_names = [];
@@ -445,7 +442,8 @@ function handle_subdirectory( string $value ) : ?array {
 	$network_url = wp_parse_url( network_site_url() );
 	$network_host = $network_url['host'];
 
-	$path = trim( $value, '/' );
+	// Break out URL query parameters and hashes.
+	$path = trim( preg_split( "/(&+|\?+|#+)/", $value )[0], '/' );
 
 	return [
 		'domain' => $network_host,
--- a/inc/add_site_ui/namespace.php
+++ b/inc/add_site_ui/namespace.php
@@ -442,7 +442,7 @@ function handle_subdomain( string $value ) : ?array {
 	$value = sanitize_domain_segment( $value );
 
 	return [
-		'domain' => $value . '.' . $network_host,
+		'domain' => idna_encode( $value . '.' . $network_host ),
 		'path' => '/',
 	];
 }
@@ -461,7 +461,7 @@ function handle_subdirectory( string $value ) : ?array {
 	$path = trim( preg_split( '/(&+|\?+|#+)/', $value )[0], '/' );
 
 	return [
-		'domain' => $network_host,
+		'domain' => idna_encode( $network_host ),
 		'path' => $path,
 	];
 }
@@ -505,7 +505,7 @@ function handle_custom_domain( string $url ) : ?array {
 	$path = trim( $url_array['path'], '/' );
 
 	return [
-		'domain' => $domain,
+		'domain' => idna_encode( $domain ),
 		'path' => '/' . $path,
 	];
 }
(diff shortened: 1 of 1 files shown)


roborourke (changes requested):
This kind of resolves the issue for subdomains and subfolders but it breaks adding custom domains. Dots are converted to hyphens resulting in invalid domain names. Additionally double hyphens are converted to a single hyphen.

This should instead just focus on sanitising input for any valid URL characters that can be used in the relevant part of a URL, `sanitize_title()` is really designed for making post titles or other string of text into a single valid URL path component so unfortunately isn't suitable here.

Things to test:

- Custom domains and subdomains with double hyphens work (eg. punycode domains) like xn--rbfka.com
- Sub directories can include dots, slashes, tilde, hyphens and underscores, and unicode characters are url encoded
- Sub sub domains can be added eg. one.two
- Query strings and hash locations are stripped eg. `?foo=bar` and `#foo`

You'll need to validate based on the type of input selected so:

- Sub folders, sanitise based on valid URL paths, stripping any query string or hash location
- Sub domains sanitise based on valid host name characters
- Custom domains, same as for sub domains but you can further validate that it's a valid hostname eg. it must have a TLD

You could probably construct a full URL and use `sanitize_url()` instead then extract the user input part of the result.

Additionally you could go a bit further and enhance the UX by adding the [`pattern` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/pattern) and changing the value using JS depending on the type of input selected although it might make it difficult to add URLs in other languages like japanese. I'm not sure the extent to which WP supports that but there's some interesting reading here https://www.w3.org/International/articles/idn-and-iri/

rmccue:
Guess what? Past life Ryan [wrote code to sanitize custom domains](https://github.com/humanmade/happytables/blob/master/content/plugins-mu/ht-domains/ht-domains.php#L88-L144) that we can reuse here :D

jazzsequence:
@rmccue There's already [a lot of very similar code](https://github.com/humanmade/altis-cms/blob/master/inc/add_site_ui/namespace.php#L188-L255) in the file itself. The thing that's missing is the `sanitize_titles_with_dashes` bit which is part of the form handler function. I'm not sure there's enough of a difference between `sanitize_titles` and `sanitize_titles_with_dashes` to change that specifically, and I think everything else should be okay, unless you see something that I'm missing.

jazzsequence:
Erp...looks like I read @rmccue's comment before @roborourke's.

jazzsequence:
Updated the original issue description with added acceptance criteria and dev notes from the above.

jazzsequence:
I have some questions about what's technically _possible_ with some of these variations but I'm dropping this here as a WIP reference of how far I've gotten.

<figure class="wp-block-table"><table><tbody><tr><td><strong>URL Pattern</strong></td><td><strong>Can create site</strong></td><td><strong>Site loads normally</strong></td></tr><tr><td>a bunch of spaces before.altis-dev.altis.dev</td><td>❌</td><td>❌</td></tr><tr><td>hyphenated-subdomain.altis-dev.altis.dev</td><td>✅</td><td>❌</td></tr><tr><td>subdomain.altis-dev.altis.dev</td><td>✅</td><td>✅</td></tr><tr><td>one.two.altis-dev.altis.dev</td><td>✅</td><td>❌</td></tr><tr><td>areallylongsinglewordsubdomain.altis-dev.altis.dev</td><td>✅</td><td>✅</td></tr><tr><td>h-y-p-h-e-n-a-t-e-d-s-u-b-d-o-m-a-i-n.altis-dev.altis.dev</td><td>✅</td><td>❌</td></tr><tr><td>two--hyphens.altis-dev.altis.dev</td><td>✅</td><td>❌</td></tr><tr><td>a custom domain with spaces.com</td><td>✅ <br>(spaces stripped)</td><td>✅</td></tr><tr><td>altis-dev.altis.dev/subdirectory with spaces</td><td>❌</td><td>❌</td></tr><tr><td>altis-dev.altis.dev/?query_string</td><td>❌</td><td>❌</td></tr><tr><td>altis-dev.altis.dev/&amp;query_param</td><td>❌</td><td>❌</td></tr><tr><td>altis-dev.altis.dev/#hash</td><td>❌</td><td>❌</td></tr><tr><td>altis-dev.altis.dev/subdirectory/</td><td>✅</td><td>✅</td></tr><tr><td>altis-dev.altis.dev/two/levels/</td><td>✅</td><td>❌</td></tr><tr><td>altis-dev.altis.dev/query?query_string</td><td>✅<br>(query string stripped)</td><td>✅</td></tr><tr><td>altis-dev.altis.dev/query-param?query&amp;param=true</td><td>✅<br>(query string stripped)</td><td>✅</td></tr><tr><td>altis-dev.altis.dev/hash#hashtag</td><td>✅<br>(hash stripped)</td><td>✅</td></tr></tbody></table></figure>

rmccue:
> altis-dev.altis.dev/two/levels/

Whether this one works or not is dependent on configuration within WP: https://github.com/WordPress/wordpress-develop/blob/3c40632397cf1ee64e09c7ac767391dfac59e5ea/src/wp-includes/ms-load.php#L166-L177

> xn--rbfka.com

Have we tested the raw input of the characters? (`Requests_IDNA` can handle encoding if we need)

jazzsequence:
@rmccue I've tried this: `JP納豆.例.jp` (taken from one of the [links @roborourke shared](https://www.w3.org/International/articles/idn-and-iri/#resolvedomain)) -- is that what you mean?

roborourke:
https://xn--rb-fka.co/ is my website aka röb.co

jazzsequence:
Thanks @roborourke I will add that to the list.

Adding this here for reference later: https://developer.wordpress.org/reference/classes/requests_idnaencoder/

jazzsequence:
With the addition of the IDNAEncoder class in the most recent commits, URL patterns like `JP納豆.jp` and `納豆.altis-dev.altis.dev` now work as expected (though they still have an nginx 404 locally). However, subdirectories (e.g. `altis-dev.altis.dev/納豆/`) throw a _WordPress_ 404. Not sure where the code responsible for this break is.

rmccue:
Note that IDNA only applies to DNS, not to paths, so we don't want/need IDNA encoding for those. I suspect those need to be URL encoded?

jazzsequence:
💥 That was the thing @rmccue. I was afraid that URL-encoding would lose the actual characters in the URL, but in testing it works as expected.

jazzsequence:
The linting errors here are coming from CSS stuff and should be addressed in a separate issue.

This specific issue seems like a good example to talk about how we can use unit testing to make the development process easier/faster and ensure consistent results. We don't have any precedent for where tests would go or how they would be run from the Altis software side, do we? (Obvs this conversation can exist outside this specific issue, it's just something I've been thinking about as I've been running through all the possible variations.)

inc/add_site_ui/namespace.php
```diff
@@ -411,11 +457,12 @@ function handle_subdirectory( string $value ) : ?array {
 	$network_url = wp_parse_url( network_site_url() );
 	$network_host = $network_url['host'];
 
-	$path = trim( $value, '/' );
+	// Break out URL query parameters and hashes.
+	$path = trim( preg_split( '/(&+|\?+|#+)/', $value )[0], '/' );
```
roborourke, line 461: I'm not sure the pluses are necessary here, and given that it's just matching on single characters you could use a character class instead eg:
```suggestion
	$path = trim( preg_split( '/[&\?#]/', $value )[0], '/' );
```

inc/add_site_ui/namespace.php
```diff
@@ -186,9 +188,46 @@ function output_add_site_page() {
  * @return bool True if valid, false otherwise.
  */
 function validate_domain_segment( string $segment ) : bool {
+	$segment = sanitize_domain_segment( $segment );
```
roborourke, line 191: I'm not keen on this being here if the function is for validating a value - it shouldn't really be modifying the value as well. I would expect the value from user input to be sanitised first and then the value of that validated.

I'm guessing that validation was failing in some cases without it so it might be that you either need to sanitize the user input early before validating or perhaps you might need to split `$segment` on `.` and validate each part of the domain.

inc/add_site_ui/namespace.php
```diff
@@ -186,9 +188,46 @@ function output_add_site_page() {
...
+ *
+ * Strips out invalid characters from a URL segment.
+ *
+ * @param string $segment Dirty segment to sanitize.
+ *
+ * @return string The cleaned segment.
+ */
+function sanitize_domain_segment( string $segment ) : string {
+	$segment = wp_kses_no_null( $segment );
+	$segment = remove_accents( $segment );
```
roborourke, line 206: Given this is passed through `idna_encode()` later why remove accents here? Was there some reason not to use `idna_encode()` at this stage?

roborourke (changes requested):
Couple of thoughts and questions here, it looks pretty close otherwise 👍

inc/add_site_ui/namespace.php
roborourke: Also the ampersand might not be necessary but doesn't do any harm.

inc/add_site_ui/namespace.php
jazzsequence: That's a good point and might fix the `röb` turning into `rob` when entered. Left it in because it was in the HappyTables stuff, but our need is probably more advanced. If we're removing that, this function seems to serve no purpose anymore, so will probably pull it out and maybe put the `wp_kses` call elsewhere.

inc/add_site_ui/namespace.php
```diff
@@ -411,11 +436,12 @@ function handle_subdirectory( string $value ) : ?array {
...
 
-	$path = trim( $value, '/' );
+	// Break out URL query parameters and hashes.
+	$path = trim( preg_split( '/[&\?#]/', $value )[0], '/' );
 
 	return [
-		'domain' => $network_host,
-		'path' => $path,
+		'domain' => idna_encode( $network_host ),
+		'path' => urlencode( $path ),
```
roborourke, line 444: Not sure why it was missing but just adding a slash prefix here so it's consistent with the other `handle_*` functions
```suggestion
		'path' => '/' . urlencode( $path ),
```

roborourke (approved):

PR #344, opened 2021-03-12
Add authorship plugin

### LINKED ISSUE #337: Integrate and test the Authorship Plugin with Altis
Authorship is a modern approach to author attribution in WordPress. It supports attributing posts to multiple authors and to guest authors, provides a great UI, and treats API access to author data as a first-class citizen.

[Authorship Plugin](https://github.com/humanmade/authorship#current-status)

**Acceptance Criteria:**
- Enable or Disable via altis-config with default off
- Integrate any settings within the altis-config
- Documentation of the plugin

roborourke:
Fixes #337

Add the plugin, the installer override, config option and docs based on the plugin's readme.

The feature itself doesn't require or need anything in the way of further integration so this is quite straightforward.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,8 @@
     "altis/cms-installer": "^0.4.3",
     "johnbillion/extended-cpts": "^4.5.1",
     "humanmade/clean-html": "^2.0.0",
-    "humanmade/post-cloner": "~1.2.0"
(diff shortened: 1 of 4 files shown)


jazzsequence (approved):
👍
Only comment here is to define CRUD which non-developers might not understand.

PR #375, opened 2021-05-03
Display user pagination in network/users.php

igmoweb:
Even if the network is considered large from within the config, display pagination when listing the network users in network/users.php

A network defined as large would prevent this and while Altis defines all networks as large by default, this was an annoying issue in the network users list.

This can be considered as a new feature and not a bug as it's a hack to a WP feature happening because of an Altis configuration field.

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -37,6 +37,14 @@ function bootstrap() {
 		add_filter( 'wp_is_large_network', function () use ( $config ) {
 			return $config['large-network'];
 		} );
+
+		if ( $config['large-network'] === true ) {
+			// Display user pagination in network/users.php even if the network is considered as large.
+			add_filter( 'users_list_table_query_args', function( $args ) {
+				$args['count_total'] = true;
+				return $args;
+			} );
+		}
 	}
 
 	if ( $config['login-logo'] ) {


roborourke:
@rmccue I think you suggested having the large network setting originally - do you remember any other specific effects it has or are you aware of any potential problems with doing this?

johnbillion:
From the perspective of someone who performed an import into an Altis site and wasted time investigating why the correct user count and pagination wasn't appearing on the Network Admin -> Users screen, and subsequently dealing with a client who reported that this screen wasn't working, the "large-network" by default is very frustrating.

The `wp_is_large_network()` setting also disables pagination on the Sites screen. If pagination is going to be re-enabled on these screens then we should revisit why `wp_is_large_network()` is set to true by default.

Here's what the `wp_is_large_network()` setting controls:

* Removes pagination and `count_total` on the Network Admin -> Users screen
* Removes pagination and `no_found_rows` on the Network Admin -> Sites screen
* Disables autocomplete when adding an existing network user to a site
* Prevents searching the `user_url` field when searching users
* Disables live update of user and site counts when adding and updating users and sites

If the live updating of user and site counts has a performance impact that needs to be managed then these can be controlled with the `enable_live_network_counts` filter.

roborourke:
We still haven't documented the `large-network` config option either so that needs to be done. How many sites that we host actually qualify as large networks? We could perhaps reverse the default setting in future but in the interim I'd be in favour of keeping the following effects:

> - Disables autocomplete when adding an existing network user to a site
> - Prevents searching the user_url field when searching users
> - Disables live update of user and site counts when adding and updating users and sites

...and reinstating pagination for users and sites in the network admin.

roborourke:
Need to follow up with another PR that documents that the config option can be set to `true`, `false` or `null` (default WP behaviour).

Additionally looks like the performance issues we want to protect against are:

- user queries having no per page limit by default
- live network count updates

PR #382, opened 2021-05-14
WordPress 5.6.4 update

jazzsequence:
Updates WordPress to the latest security patch.
https://wordpress.org/support/wordpress-version/version-5-6-4/


roborourke (approved):
Need to fix dependabot for these, we missed 5.6.3 from the look of it.

jazzsequence:
Yeah, I think Dependabot isn't looking at the `v*-branch`es

PR #391, opened 2021-05-31
Add altis css color variables

jazzsequence:
Adds global color CSS variables for Altis colors per the Altis brand guidelines.

related to work in humanmade/altis-seo#53 and humanmade/altis-seo#54

### DIFF
--- a/assets/admin-color-scheme.css
+++ b/assets/admin-color-scheme.css
@@ -9,9 +9,25 @@
 /* stylelint-disable declaration-no-important, max-line-length, no-descending-specificity, selector-class-pattern */
 
 :root {
-	--wp-admin-theme-color: #4667de;
-	--wp-admin-theme-color-darker-10: #2041ab;
-	--wp-admin-theme-color-darker-20: #183080;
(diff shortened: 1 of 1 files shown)


roborourke (approved):
Does WP admin support IE11? Thinking not, but this does have some bearing on this change. The updates looks good though 👍

jazzsequence:
> Does WP admin support IE11?

WordPress has their own color CSS variables (we override them in our theme), so I thought about this, but assumed if they were using them, then it was fair game. [This Make Core post](https://wordpress.org/news/2021/05/dropping-support-for-internet-explorer-11/) (and some various reports on the internets) confirms they will officially drop support for IE11 in 5.8 (which aligns with Altis v8). So this shouldn't be backported but will be fine for v8 release.

jazzsequence:
In addition, it looks like Yoast SEO [dropped support for IE11 in 12.6](https://github.com/Yoast/wordpress-seo/issues/11809#issuecomment-558977314), so we would kind of need to also drop support if we're including Yoast.

PR #406, opened 2021-07-19
FIx code comment in documentation for `simple_local_avatars_upload_limit` filter

goldenapples:
This filter is operating on bytes, not kilobytes, so `2 * 1024` is 2KB, not 2MB.

### DIFF
--- a/docs/local-avatars.md
+++ b/docs/local-avatars.md
@@ -44,6 +44,6 @@ Allows overriding of the maximum allowable file size for avatar uploads. Default
 
 ```php
 add_filter( 'simple_local_avatars_upload_limit', function() {
-	return 2 * 1024; // Max 2MB.
+	return 2 * 1024; // Max 2KB.
 } );
 ```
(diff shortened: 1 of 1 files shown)


roborourke:
Hmm, we probably actually want it to be 2Mb, 2kb for an image is miniscule.

roborourke:
Oh sorry that's the docs. Still, I think the example probably wants to reflect MB

docs/local-avatars.md
```diff
@@ -44,6 +44,6 @@ Allows overriding of the maximum allowable file size for avatar uploads. Default
 
 ```php
 add_filter( 'simple_local_avatars_upload_limit', function() {
-	return 2 * 1024; // Max 2MB.
+	return 2 * 1024; // Max 2KB.
```
roborourke, line 47:
```suggestion
	return 2 * 1024 * 1024; // Max 2MB.
```

roborourke (approved):

PR #412, opened 2021-07-22
Update wp-config.php

tfrommen:
Fix comment, use `is_readable` instead of `file_exists`, and adapt file loading.


roborourke:
I get that `is_readable()` is a better function to use but what is the benefit of using `include_once` over `require_once`?

roborourke (approved):

tfrommen:
@roborourke the main difference between `require` and `include` is that the former will error if the file cannot be loaded. Which will never be the case here, given we checked its existence **and** readability.

There's also a PHPCS sniff specific to these "guarded includes": https://github.com/squizlabs/PHP_CodeSniffer/blob/master/src/Standards/PEAR/Sniffs/Files/IncludingFileSniff.php

roborourke:
Ah of course, thanks for the clarification.

PR #414, opened 2021-08-03
Fix logic error for defining WP_INITIAL_INSTALL

mattheu:
@roborourke  I don't think this code does quite what you intended. The comment for this condition suggests this should return true for PHP Unit and direct calls to `install.php`. 

However your condition uses `&&` which is the same, I think you mean `||`.

I should also note that `php_sapi_name() === 'cli'` could return true in other circumstances such as when running WP-CLI commands, or WP-Shell, so I'm not convinced this is a robust solution for this issue.


roborourke:
This doesn't seem right as it would make the function return `true` for any CLI command. We're specifically interested only in direct CLI invocations of `install.php` there which is what the WP PHPUnit tools do.

mattheu:
OK I think I'm heading in the wrong direction here.

PR #425, opened 2021-09-09
Add v8-branch to dependabot

rmccue:


rmccue:
cc @roborourke we probably need to do these somewhere else too

kadamwhite (approved):
Approvéd

PR #437, opened 2021-09-23
Network enable themes by default

### LINKED ISSUE #14: Network Enable all themes by default
Having to enable new themes in the Network Admin doesn't seem like it's really giving us anything useful. The use case so far as I can tell is for sites like WP.com.

Should we just enable all themes by default instead?

-----------

Acceptance criteria:

- [x] All themes are network enabled by default
- [x] A config option can be set to toggle this feature, defaults to on
- [x] Config option is documented

roborourke:
Fixes #14

### DIFF
--- a/docs/themes.md
+++ b/docs/themes.md
@@ -24,25 +24,3 @@ When a [new site](docs://guides/multiple-sites.md) is created, a "default" theme
 	}
 }
 ```
-
-## Network Enabled Themes
-
-By default all themes are network enabled, this means that every theme in your application is available to all sites on the network.
-
-Individual themes can still be disabled for the network, and enabled for specific sites by editing a site in the network admin.
-
-If you prefer to use the standard WordPress behaviour where themes are disabled for the network by default you can set the `network-enable-themes` config option to `false`:
-
-```json
-{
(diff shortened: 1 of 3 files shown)


docs/themes.md
```diff
@@ -24,3 +24,25 @@ When a [new site](docs://guides/multiple-sites.md) is created, a "default" theme
...
+
+If you prefer to use the standard WordPress behaviour where themes are disabled for the network by default you can set the `network-enable-themes` config option to `false`:
+
+```json
+{
+	"extra": {
+		"altis": {
+			"modules": {
+				"cms": {
+					"network-enable-themes": false
```
rmccue, line 42: I wonder if something like `enable-all-themes` might be clearer?

roborourke: I'm not sure there is a clear name for it given that you subsequently disable them. Nothing is being set in stone so it doesn't really feel like a config option is actually a good fit to be honest.

rmccue: hmm, yeah; `enable-new-themes-by-default` which is a bit long winded 🤔

shadyvb (approved):

PR #443, opened 2021-10-15
Add a note on using blog prefix for permalinks

shadyvb:
Note that the filters in question are not limited to the main site, so we should add a check for is_multisite and is_main_site to make this usable for non-main-sites as well.

Related https://github.com/humanmade/altis-cms/issues/70

### DIFF
--- a/docs/permalinks.md
+++ b/docs/permalinks.md
@@ -1,0 +1,10 @@
+# Permalink
+
+Altis introduces a fix to a bug where the main site on a network is hardcoded to contain `/blog/` prefix for posts by default. The side-effect for this is that you cannot use `/blog/` prefix on the main site unless you disable the fix, which is by unhooking the filters associated with it, eg:
+
+```php
+remove_filter( 'sanitize_option_permalink_structure', '\Altis\CMS\Permalinks\remove_blog_prefix' );
+remove_filter( 'option_permalink_structure', '\Altis\CMS\Permalinks\remove_blog_prefix' );
+```
+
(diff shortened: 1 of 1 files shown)


roborourke:
Think my comment was based on the opening lines of the docs regarding it being a bug.

I think I'll add a new issue to document permalinks more generally, along with the capabilities extended custom post types and taxonomies adds.

docs/permalinks.md
```diff
@@ -0,0 +1,10 @@
+# Permalink
```
roborourke, line 1:
```suggestion
# Permalinks
```

docs/permalinks.md
```diff
@@ -0,0 +1,10 @@
...
+
+Altis introduces a fix to a bug where the main site on a network is hardcoded to contain `/blog/` prefix for posts by default. The side-effect for this is that you cannot use `/blog/` prefix on the main site unless you disable the fix, which is by unhooking the filters associated with it, eg:
+
+```php
+remove_filter( 'sanitize_option_permalink_structure', '\Altis\CMS\Permalinks\remove_blog_prefix' );
+remove_filter( 'option_permalink_structure', '\Altis\CMS\Permalinks\remove_blog_prefix' );
```
roborourke, line 7: I think it'll save us some potential support to wrap these in an action so we know they're being removed after they've been added, I think `plugins_loaded` would be fine:

```suggestion
add_action( 'plugins_loaded', function () {
    remove_filter( 'sanitize_option_permalink_structure', '\Altis\CMS\Permalinks\remove_blog_prefix' );
    remove_filter( 'option_permalink_structure', '\Altis\CMS\Permalinks\remove_blog_prefix' );
} );
```

roborourke (approved):

Issue #207, opened 2020-03-19
Add Advanced Post Cache plugin


roborourke:
Add https://github.com/Automattic/advanced-post-cache

This is blocked by a8c not having published the package to packagist yet.

Acceptance criteria:

- [ ] Advanced post cache is required in `composer,json`
- [ ] Advanced post cache file is loaded on the `muplugins_loaded` hook in `inc/namespace.php`
- [ ] The Advanced post cache package is referenced in the `install-overrides` array in `composer.json`

PriyaBhatia0210:
Blocked by upstream

rmccue:
Looks like this is unblocked: https://packagist.org/packages/automattic/advanced-post-cache

robindevitt:
> Above plugin are added to the exclude list in humanmade/altis-core here https://github.com/humanmade/altis-core/blob/master/inc/composer/class-installer.php#L26

Should this be the correct URL : https://github.com/humanmade/altis-core/blob/master/inc/composer/class-override-installer.php ?

roborourke:
@robindevitt that's now a legacy list, you need to add the package name to the list in the `installer-overrides` section of this repo's `composer.json` instead. I'll get those docs updated.

robindevitt:
@roborourke From your [comment in slack](https://hmn.slack.com/archives/C5WSNGC9F/p1636116975164900?thread_ts=1636106496.144100&cid=C5WSNGC9F) around these lines https://github.com/humanmade/altis-cms/blob/master/composer.json#L43-L48

1. Are those strings relevant to the GitHub URL's or to a folder for each item?
2. I have added a string for `"Automattic/advanced-post-cache"` inside of `install-overrides`, is there anything I need to "run" to get it to install so I will see the field in the vendor folder?

roborourke:
@robindevitt 

> Are those strings relevant to the GitHub URL's or to a folder for each item?

They represent the composer package name which is set by the `name` property in a given repo's `composer.json` file, in this case it is `automattic/advanced-post-cache`: https://github.com/Automattic/advanced-post-cache/blob/master/composer.json#L2

Note that it is lower case.

> I have added a string for "Automattic/advanced-post-cache" inside of install-overrides, is there anything I need to "run" to get it to install so I will see the plugin in the vendor folder?

[...]

PR #468, opened 2021-11-08
Add advanced post cache plugin

robindevitt:
Add the advanced post cache plugin via composer. #207

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
     "humanmade/altis-reusable-blocks": "~0.2.0",
     "humanmade/asset-loader": "^0.5.0",
     "10up/simple-local-avatars": "~2.2.0",
-    "automattic/advanced-post-cache": "dev-master"
+    "automattic/advanced-post-cache": "dev-master#c46b2d95773689938ef1e89a3dfcb9cbd94022b8"
   },
   "autoload": {
     "files": [
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,8 @@
     "humanmade/authorship": "~0.2.9",
     "humanmade/altis-reusable-blocks": "~0.2.0",
     "humanmade/asset-loader": "^0.5.0",
-    "10up/simple-local-avatars": "~2.2.0"
+    "10up/simple-local-avatars": "~2.2.0",
+    "automattic/advanced-post-cache": "dev-master"
   },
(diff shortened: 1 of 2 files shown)


roborourke:
When you add a PR you should always reference the related issue(s) in the PR comment rather than the PR title, that way we'll get the cross-linking / referencing.

A brief description of the change is always helpful too.

composer.json
```diff
@@ -19,7 +19,8 @@
     "humanmade/authorship": "~0.2.9",
     "humanmade/altis-reusable-blocks": "~0.2.0",
     "humanmade/asset-loader": "^0.5.0",
-    "10up/simple-local-avatars": "~2.2.0"
+    "10up/simple-local-avatars": "~2.2.0",
+    "automattic/advanced-post-cache": "dev-master"
```
roborourke, line 23: Only change to make here is to pin a specific commit. If we use `dev-master` then it will always fetch the latest version from the `master` branch which can change outside of our control. For stability we always want to install a known or at least trusted state of the code, this is why typically we'll use tags only, or tags with a version constraint like `^` or `~`.

```suggestion
    "automattic/advanced-post-cache": "dev-master#c46b2d95773689938ef1e89a3dfcb9cbd94022b8"
```

roborourke (changes requested):
Looks spot on barring the one suggested change. I'd test with that just to confirm.

composer.json
robindevitt: Thank you, [Resolved in this commit](https://github.com/humanmade/altis-cms/pull/468/commits/cf650b6ceba3ebbacc9c71cd997af710bd5af603)

roborourke:
Looked into this a bit further and this won't install successfully as is. We need the upstream to create a tagged release as a stable version. Stability flag like `@dev` are only supported in a project root `composer.json`

robindevitt:
I have submit an issue on the [Advanced Post Cache repo](https://github.com/Automattic/advanced-post-cache/issues/25) to request a tag in order for us to make use of it. Will monitor it for any further progress.

kovshenin:
Query cache is included in WordPress 6.1: https://core.trac.wordpress.org/ticket/22176

kovshenin:
Closing issue in favor of core's implementation.

PR #479, opened 2021-11-22
[tests] CMS module QA tests

shadyvb:
Fixes https://github.com/humanmade/product-dev/issues/834

Needs https://github.com/humanmade/altis-dev-tools-command/pull/18

### DIFF
--- a/tests/acceptance/AuthorshipCest.php
+++ b/tests/acceptance/AuthorshipCest.php
@@ -13,6 +13,13 @@ use Codeception\Util\Locator;
 class AuthorshipCest {
 
 	/**
+	 * Rollback callback for the authorship activation bootstrap call.
+	 *
+	 * @var callable
+	 */
+	protected $rollback = null;
+
+	/**
 	 * Make sure Authorship is activated.
 	 *
 	 * @param AcceptanceTester $I Actor object.
@@ -20,7 +27,19 @@ class AuthorshipCest {
 	 * @return void
 	 */
 	public function _before( AcceptanceTester $I ) {
-		$I->bootstrapWith( [ __CLASS__, '_enableAuthorship' ] );
+		$this->rollback = $I->bootstrapWith( [ __CLASS__, '_enableAuthorship' ] );
+	}
+
+	/**
+	 * Deactivate authorship after tests are finished.
+	 *
+	 * @param AcceptanceTester $I Actor.
+	 *
+	 * @return void
+	 */
+	public function _after( AcceptanceTester $I ) {
+		$callback = $this->rollback;
+		$callback();
 	}
 
 	/**
--- a/tests/acceptance/AuthorshipCest.php
+++ b/tests/acceptance/AuthorshipCest.php
@@ -1,0 +1,135 @@
+<?php
+/**
+ * Tests for core module's admin features.
+ *
+ * phpcs:disable WordPress.Files, WordPress.NamingConventions, PSR1.Classes.ClassDeclaration.MissingNamespace, HM.Functions.NamespacedFunctions
+ */
+
+use Codeception\Util\Locator;
+
+/**
+ * Test core module admin features.
+ */
+class AuthorshipCest {
+
+	/**
+	 * Make sure Authorship is activated.
+	 *
+	 * @param AcceptanceTester $I Actor object.
+	 *
+	 * @return void
+	 */
+	public function _before( AcceptanceTester $I ) {
+		$I->bootstrapWith( [ __CLASS__, '_enableAuthorship' ] );
+	}
+
+	/**
+	 * Activate authorship.
+	 *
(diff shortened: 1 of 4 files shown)


tests/acceptance/AuthorshipCest.php
```diff
@@ -0,0 +1,135 @@
...
+
+	/**
+	 * Test creating a new post with both a guest and an author user.
+	 *
+	 * @param AcceptanceTester $I Tester
+	 *
+	 * @return void
+	 */
+	public function testAuthorship( AcceptanceTester $I ) {
+		$I->haveUserInDatabase( 'Arthur', 'author' );
```
roborourke, line 46: Ha! I always use this name for test author accounts too :)

tests/acceptance/AuthorshipCest.php
```diff
@@ -0,0 +1,135 @@
...
+		$post_id = $I->grabFromDatabase( $I->grabPostsTableName(), 'ID', [ 'post_title' => 'Test post 1' ] );
+		$guest_user_id = $I->grabUserIdFromDatabase( 'Gusto' );
+		$author_user_id = $I->grabUserIdFromDatabase( 'Arthur' );
+		$guest_term_id = $I->grabTermIdFromDatabase( [ 'slug' => $guest_user_id ] );
+		$author_term_id = $I->grabTermIdFromDatabase( [ 'slug' => $author_user_id ] );
+		$I->seePostWithTermInDatabase( $post_id, $guest_term_id, null, 'authorship' );
+		$I->seePostWithTermInDatabase( $post_id, $author_term_id, null, 'authorship' );
+
+		// Test post API response has expected authorship values.
+		$I->amOnPage( '/wp-json/wp/v2/posts/' . $post_id );
```
roborourke, line 101: 🤯

tests/acceptance/AuthorshipCest.php
```diff
@@ -0,0 +1,135 @@
...
+		$source = $I->grabTextFrom( 'pre' );
```
roborourke, line 102: Is this perhaps browser dependent? I'm not sure if there's a uniform way browsers render JSON responses. Maybe it could be `<body>`?

roborourke (commented):
This looks good so far. We can split up the issue further if you like to bring items into the next sprint.

tests/acceptance/AuthorshipCest.php
shadyvb: This will need a helper method, for a future ticket.

tests/acceptance/AuthorshipCest.php
shadyvb: Yeah, will need to test with Firefox. Good catch!

tests/acceptance/AuthorshipCest.php
```diff
@@ -0,0 +1,135 @@
...
+
+	/**
+	 * Make sure Authorship is activated.
+	 *
+	 * @param AcceptanceTester $I Actor object.
+	 *
+	 * @return void
+	 */
+	public function _before( AcceptanceTester $I ) {
+		$I->bootstrapWith( [ __CLASS__, '_enableAuthorship' ] );
```
roborourke, line 23: Does this need rollback set to false? As I understand it the file will get cleaned up as soon as the headless browser navigates somewhere

roborourke (approved):

tests/acceptance/AuthorshipCest.php
shadyvb: Nope, as soon as the test thread closes, not the acceptance test thread. I should actually rollback after the test.

roborourke: Ahhh of course 🤦‍♂️  - pretty complicated thinking between what's in the test thread and what's in the browser request thread! All LGTM :shipit:

tests/acceptance/AuthorshipCest.php
```diff
@@ -0,0 +1,153 @@
...
+
+	/**
+	 * Make sure Authorship is activated.
+	 *
+	 * @param AcceptanceTester $I Actor object.
+	 *
+	 * @return void
+	 */
+	public function _before( AcceptanceTester $I ) {
+		$this->rollback = $I->bootstrapWith( [ __CLASS__, '_enableAuthorship' ] );
```
roborourke, line 30: Idea for a later addition, maybe we could add this to the helper where `bootstrapWith` is defined, so you could do a class level bootstrap as well as individual test level.

roborourke (approved):

tests/acceptance/AuthorshipCest.php
shadyvb: Not sure I'm following, can you elaborate a bit more ? Tests currently does not extend any class, so not sure which class you're referring to.

Issue #112, opened 2019-10-11
No Favicon in WP Admin


tomjn:
<img width="71" alt="Screenshot 2019-10-11 at 16 29 28" src="https://user-images.githubusercontent.com/58855/66664373-50db6e80-ec44-11e9-81da-dffcfc894181.png">
WP Admin and the frontend have no favicon by default, there should be a default favicon with an Altis logo for the back end

---------

Acceptance criteria:

- [ ] Favicon defaults to Altis favicon if not set

rmccue:
Is there a reason we should include this? This is without a doubt going to be customised on every site, so not sure we should bother.

tomjn:
Oh I agree, especially on the frontend, but it's a polish thing. When you first fire up a local environment it has the default which is suboptimal.

We should be able to plug into cores site icon functionality and provide an altis logo as a fallback for when nothing has been specified in the admin area. I don't know that i'd be comfortable doing that on the frontend though

rmccue:
Should we make the favicon in the admin always be the Altis logo? @grarighe (cc @noeltock)

noeltock:
@Camila-villegas When you get a chance, can you throw something quick together?
- Start with 512x512
- Requires transparency
- Can test here https://favicon.io/favicon-converter/

Thank you!

joehoyle:
Could we just use the one from altis-dxp.com ?

Camila-villegas:
Forwarding you guys the one we have been using:  https://drive.google.com/open?id=1-2oIQ3ZNy6UvAC9xF6aqwwJwc2llkahM

robindevitt:
Similar to how the login logo is set, should the favicon be managed through the the `config` file too?

robindevitt:
[PR](https://github.com/humanmade/altis-cms/pull/483) is ready for review, requested from Rob or Shadi.

robindevitt:
Rob got a couple changes done and closed the issue out.

PR #483, opened 2021-12-30
No Favicon in WP Admin

robindevitt:
In the WP Admin and the frontend have no favicon by default, there should be a default favicon with an Altis logo for the back end.

### Solution
The Altis Module adds a default Altis favicon and gives the user the ability to set the favicon from the config file. Setting the favicon from the customizer will override the favicon set in the config.

Issue #112


inc/namespace.php
```diff
@@ -51,6 +51,12 @@ function bootstrap() {
 		add_action( 'login_header', __NAMESPACE__ . '\\add_login_logo' );
 	}
 
+	if ( isset( $config['favicon-icon'] ) ) {
```
shadyvb, line 54: Should we allow devs to override the behavior by using `! empty( .. )` instead ? So they'd be able to use false and empty strings as values.

shadyvb: Actually, could we use the filter in `get_site_icon_url` instead ? So we don't have to output our own markup, and ensure consistency with WordPress markup ?

Ref: https://developer.wordpress.org/reference/functions/get_site_icon_url/

shadyvb (changes requested):

inc/namespace.php
robindevitt: What if I were to change it to use `has_site_icon` instead so if the site Icon is set already, then those functions don't even run yet.

Ref: https://developer.wordpress.org/reference/functions/has_site_icon/

With that change it would eliminate the functions from running unnecessary. As the `favicon-icon` has a default setting so it would always essentially have a value.

shadyvb: Hmmm, possibly both ? Checking if there's a valid config value OR a valid setting value as well. Then hooking to the filter.

inc/namespace.php
```diff
@@ -51,6 +51,12 @@ function bootstrap() {
 		add_action( 'login_header', __NAMESPACE__ . '\\add_login_logo' );
 	}
 
+	if ( ! empty( $config['favicon-icon'] ) ) {
```
robindevitt, line 54: @shadyvb I couldn't do a check here for `has_site_icon`,  it would present a "Fatal Error" so I went with your suggestions check if the value was not empty.

inc/namespace.php
```diff
@@ -267,6 +273,18 @@ function add_login_logo() {
...
+ * Add the favicon-icon when the site has no icon set.
+ */
+function add_favicon_icon() {
+	if( has_site_icon() ){
+		return;
+	}
+
+	$favicon = get_site_icon_url( '', Altis\get_config()['modules']['cms']['favicon-icon'] );
+	echo '<link rel="icon" href="' . esc_url( $favicon ) . '" />';
+}
```
robindevitt, line 286: @shadyvb I included a check here for when the `has_site_icon` and returns if it's present before continuing for use the filter you suggested.

inc/namespace.php
shadyvb: That works, you can short circuit within the filter if needed. The possibly-redundant function calls shouldn't be a big concern.

inc/namespace.php
shadyvb: I'd get rid of `add_favicon_icon` altogether and just use the filter to let WordPress do its markup output. Sorry if i wasn't clear on that.

roborourke:
One general observation - `favicon-icon` is a bit redundant so just `favicon` would be fine for the config option.

I'd also make sure that the URL is fully qualified when it's output e.g. `https://example.com/vendor/...`

robindevitt:
Hey @shadyvb 

Thank you for the input and direction so far,

I was able to re-request a review earlier but can't seem to get it now. If you have a moment to please re-review the changes I have implemented and let me know if this is better.

shadyvb (approved):
Great work!

inc/namespace.php
```diff
@@ -267,6 +271,22 @@ function add_login_logo() {
 	<?php
 }
 
+/**
+ * Undocumented function
```
roborourke, line 275: 😛

inc/namespace.php
```diff
@@ -267,6 +271,22 @@ function add_login_logo() {
...
+ *
+ * @param string $url Filters the site icon URL.
+ *
```
roborourke, line 278: No need for the empty line here.
```suggestion
```

inc/namespace.php
```diff
@@ -267,6 +271,22 @@ function add_login_logo() {
...
+ * @return string
+ */
+function filter_favicon( $url ) {
+	if ( ! $url ) {
```
roborourke, line 282: I think it's best to use `empty( $url )` instead of `! $url` to be more explicit. You could also add type hints here at least for the return value.

docs/branding.md
```diff
@@ -54,3 +54,30 @@ function add_colour_scheme() {
...
+#### Configuration option
+You can use the `favicon-icon` configuration option to provide a project root relative path to an image file.
+
+```json
+{
+	"extra": {
+		"altis": {
+			"modules": {
+				"cms": {
+					"favicon-icon": "/content/favicon.png"
```
roborourke, line 77:
```suggestion
					"favicon": "/content/favicon.png"
```

docs/branding.md
```diff
@@ -54,3 +54,30 @@ function add_colour_scheme() {
...
+
+#### CMS option
+1. Login to the admin area
+2. Go to <strong>Appearance</strong> > <strong>Customize</strong> and click on the <strong>Site Identity</strong> tab
+3. Scroll to <strong>Site Icon</strong> click on the select site icon button and then upload the image you want to use as the icon.
+4. Make any adjustment to the image if it's too large.
+5. Publish your changes.
+
+#### Configuration option
+You can use the `favicon-icon` configuration option to provide a project root relative path to an image file.
```
roborourke, line 69:
```suggestion
You can use the `favicon` configuration option to provide a project root relative path to an image file.
```

docs/branding.md
```diff
@@ -54,3 +54,30 @@ function add_colour_scheme() {
...
 }
 ```
+
+## Custom Favicon
+The favicon shown on the CMS can be set using wither the CMS or Configuration option. Using the CMS option will override the Configuration option.
+
+#### CMS option
+1. Login to the admin area
+2. Go to <strong>Appearance</strong> > <strong>Customize</strong> and click on the <strong>Site Identity</strong> tab
+3. Scroll to <strong>Site Icon</strong> click on the select site icon button and then upload the image you want to use as the icon.
```
roborourke, line 64: HTML in markdown is usually a last resort for unsupported or complex markup, use `**` to denote bold text instead.

https://www.markdownguide.org/basic-syntax/

docs/branding.md
```diff
@@ -54,3 +54,30 @@ function add_colour_scheme() {
 	);
 }
 ```
+
+## Custom Favicon
+The favicon shown on the CMS can be set using wither the CMS or Configuration option. Using the CMS option will override the Configuration option.
```
roborourke, line 59:
```suggestion
The favicon shown for the application can be set using either the CMS or a configuration option. Using the CMS option will override the configuration option.
```

docs/branding.md
```diff
@@ -54,3 +54,30 @@ function add_colour_scheme() {
...
+
+#### CMS option
```
roborourke, line 61: The previous heading level was 2 so this should be 3:
```suggestion
### CMS Option
```

docs/branding.md
```diff
@@ -54,3 +54,30 @@ function add_colour_scheme() {
...
+1. Login to the admin area
+2. Go to <strong>Appearance</strong> > <strong>Customize</strong> and click on the <strong>Site Identity</strong> tab
+3. Scroll to <strong>Site Icon</strong> click on the select site icon button and then upload the image you want to use as the icon.
+4. Make any adjustment to the image if it's too large.
+5. Publish your changes.
+
+#### Configuration option
```
roborourke, line 68:
```suggestion
### Configuration Option
```

roborourke (changes requested):
Some changes on the polish front here, it's good to sweat the small stuff too to deliver a quality product.

inc/namespace.php
robindevitt: 🙈  It's one of those at the time of filling it out I started on the parameters and would "come back" to the last change!

FWIW I found these also in a search, not sure if it's worth going through and changing them too? ( obviously one if mine that's been changed now )
Screenshot 2022-01-04 at 08 31 13

roborourke: Heh, classic! Can fix the cloud module one but the others aren’t too important.

inc/namespace.php
```diff
@@ -267,6 +271,21 @@ function add_login_logo() {
...
+/**
+ * Filter to set the favicon url.
+ *
+ * @param string $url Filters the site icon URL.
+ * @return string $url The URL for the favicon.
+ */
+function filter_favicon( string $url ) {
+	if ( empty( $url ) ) {
+		$favicon = Altis\get_config()['modules']['cms']['favicon'];
+		$url = get_site_url( get_main_site_id( get_main_network_id() ), $favicon );
```
roborourke, line 283: Oh actually - this is a heavy function to run as it involves `switch_to_blog()` on any non primary site. We have this function in the Cloud module that could be good to put in the core module, for now though I think we just need to cache the value set from the config to avoid the `get_site_url()` call after the first run:

https://github.com/humanmade/altis-cloud/blob/7b04b086ab37bc2156647649c7f6a26d73ce9c51/inc/namespace.php#L541-L555

docs/branding.md
```diff
@@ -54,3 +54,30 @@ function add_colour_scheme() {
...
 ```
+
+## Custom Favicon
+The favicon shown for the application can be set using either the CMS or a configuration option. Using the CMS option will override the configuration option.
+
+### CMS Option
+1. Login to the admin area
+2. Go to **Appearance** > **Customize** and click on the **Site Identity** tab
+3. Scroll to **Site Icon** click on the select site icon button and then upload the image you want to use as the icon.
+4. Make any adjustment to the image if it's too large.
```
roborourke, line 65:
```suggestion
3. Scroll to **Site Icon** and click on the select site icon button, then upload the image you want to use as the icon.
4. Make any adjustments needed to the image if it's too large.
```

inc/namespace.php
```diff
@@ -267,6 +271,21 @@ function add_login_logo() {
...
+	}
```
roborourke, line 284:
```suggestion
	if ( empty( $url ) ) {
		$favicon = Altis\get_config()['modules']['cms']['favicon'];
		$url = Altis\Cloud\get_main_site_url( $favicon );
	}
```

roborourke (approved):

PR #494, opened 2022-01-19
Load plugins after initial install

roborourke:
The signups plugin will try to create or upgrade its db tables on `admin_init` now. Loading these was throwing up errors during the installation process.


robindevitt:
@roborourke there a couple HM Linter checks that failed. not sure if they need to be sorted before completing the review?

roborourke:
Yeah they definitely do, and this repo has tests so those need to be passing first as well.

robindevitt (approved):

roborourke:
Ah crap, missed updating the function name in the `add_action()` call 🤦

Issue #359, opened 2021-04-06
Increase the admin toolbar hover contrast


johnbillion:
When hovering over an admin toolbar menu item, the change in colour is barely noticeable. It changes from very light grey to white, whereas vanilla WP changes from grey to blue on hover.

It would be good if this hover colour had greater contrast with the link colour.

## Altis

<img width="142" alt="" src="https://user-images.githubusercontent.com/208434/113710886-90253700-96e4-11eb-8239-f1c4f48830ef.png">

## Vanilla WP

<img width="152" alt="" src="https://user-images.githubusercontent.com/208434/113710957-a7642480-96e4-11eb-86f6-6436785f1393.png">

roborourke:
Fixed in #506

PR #506, opened 2022-01-27
Increase the admin toolbar hover contrast.

robindevitt:
### Issue
#359 When hovering over an admin toolbar menu item, the change in colour is barely noticeable. It changes from very light grey to white. ( Hovering over Media )

current-state

### Solution
Changed the menu item color from that of the sidebar links to keep some consistency. ( Hovering over Media )

new-state

### DIFF
--- a/assets/admin-color-scheme.css
+++ b/assets/admin-color-scheme.css
@@ -324,7 +324,8 @@ ul#adminmenu > li.current > a.current::after {
 #wpadminbar .quicklinks .menupop ul li a,
 #wpadminbar .quicklinks .menupop.hover ul li a,
 #wpadminbar.nojs .quicklinks .menupop:hover ul li a {
-	color: #a2aabe;
+	color: #e2ecf1;
+	border-left: 3px solid transparent;
 }
 
 #wpadminbar .quicklinks li .blavatar,
@@ -351,6 +352,10 @@ ul#adminmenu > li.current > a.current::after {
 #wpadminbar li:hover #adminbarsearch::before,
 #wpadminbar li #adminbarsearch.adminbar-focused::before {
 	color: #fff;
+}
+
+#wpadminbar .quicklinks .menupop.hover ul li > a:hover {
+	border-left-color: #fff;
 }
 
 #wpadminbar .quicklinks li a:hover .blavatar,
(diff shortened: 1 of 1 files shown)


roborourke:
Does this work ok with sub sub menus too? My concern is the reduction in contrast for all the other text in the menu. Can you check it in https://webaim.org/resources/contrastchecker/

Should be WCAC AA compliant.

Maybe we can play with the background colour or border highlighting like the main left hand admin menu

robindevitt:
It passes all the WCAC AA checks. ( new tool I learnt! ) I also learnt there is a contrast checker built into chrome browser tools :


I see I missed the lighter Altis sub-menu, which I'll look into. ( messy screengrab, but it shows all the menus with some overlapping )

roborourke:
Cheers for checking Robin, so it does track with the left hand menu now. Could we explore the left hand boarder indicator like the left hand admin menu has a further improvement? I can foresee it being a bit of pain so don't spend too long on it.

Approving this as is at any rate 👍

roborourke (approved):

roborourke:
Failing test is unrelated but should be looked into, just noting here as not a blocker for this PR.

robindevitt:
@roborourke , thoughts on keeping the original low contrast and then add the white border? I'm battling to get suitable colors within the Altis options to pass all the WCAC checks. 

Screenshot 2022-01-31 at 10 23 11
Screenshot 2022-01-31 at 10 23 23

roborourke:
Yeah if it's what's there right now then we can go for the lower contrast same as the left hand menu.

I'd like to revisit at some point as it does feel like the non hovered text is a bit too dark. FWIW if the Altis colour scheme is missing a value that would look visually better and solve a problem, such as a lighter grey then it's ok to add one in.

robindevitt:
Hey @roborourke 

I have pushed up the latest change with the original colors and added the border on hover. I feel it's better than before.

> Failing test is unrelated but should be looked into, just noting here as not a blocker for this PR.

Can I merge this and close it out with the failed check? Should I then create a new bug issue to follow it up and add it to the sprint?

roborourke:
Yeah, the failed check looks like it's to do with Authorship so good to merge this.

Issue #482, opened 2021-12-22
Enqueued scripts


robindevitt:
Steps to reproduce:
1. Login to the site
2. Either view the widgets page or view widgets in the customiser
3. View query Monitor for the following error:
> wp_enqueue_script() was called incorrectly. "wp-editor" script should not be enqueued together with the new widgets editor (wp-edit-widgets or wp-customize-widgets). Please see Debugging in WordPress for more information. (This message was added in version 5.8.0.)

Expected to see:
No Error

roborourke:
Need to list out all plugins that use this and carry out necessary refactoring, making reliance on `wp-editor` optional or conditional depending on context.

- ElasticPress
- AWS Analytics - Experience Blocks
- Altis CMS
- Reusable blocks
- Yoast SEO

robindevitt:
Starting to look into this and it appears that Yoast SEO had changes done and the only issued plugins are

- ElasticPress
- AWS Analytics - Experience Blocks
- Altis CMS
- Reusable blocks

robindevitt:
I have dug through this and starting to pull my hair out a little.

I have figured I need to focus on a single module at a time as that makes the most amount of sense and once the solution is figured out for the one, it's easier to roll out a fix for the rest.

I have made the change manually to ElasticPress until I have the solution and can then submit an issue to the original repo.

I have gone and disabled the 3 modules : `workflow`, `cms` and `analytics`.

Viewing the widgets screen I no longer see the enqueued script warning for `wp-editor`.

I have then enabled the `cms` module as this module has a single instance of `wp-editor` and confirm the error returns when the module is enabled.

I think I was being pretty hopeful thinking to remove the `wp-editor` from the wp_enqueue_script array of dependants within the module would "solve" the issue to come degree, but no luck there, the issue continues to display.

Within Query Monitor it has the following for showing the error, may I be looking in the wrong spot initially?

![Screenshot 2022-02-15 at 10.13.51.png](https://images.zenhubusercontent.com/615d863972ceea502ec6a54f/6ac40a0b-5738-4fe8-a7f1-4e7ce41f7738)

robindevitt:
Figured out cloning some of the modules into the `extra-packages` folder and then running `composer symlink` didn't link all the folders and noticeably the one I was working in

roborourke:
[...]

PR #531, opened 2022-02-21
Enqueued script warning for wp-editor on widgets screen

robindevitt:
This PR fixes the original issue reported in #482 where the below error shown when on the widgets page:

> wp_enqueue_script() was called incorrectly. "wp-editor" script should not be enqueued together with the new widgets editor (wp-edit-widgets or wp-customize-widgets). Please see Debugging in WordPress for more information. (This message was added in version 5.8.0.)

Steps to confirm:

1. Login to the site
2. View the widgets page
3. View query Monitor, you shouldn't see the error mentioned above that `wp_enqueue_script() was called incorrectly.`

### DIFF
--- a/inc/branding/namespace.php
+++ b/inc/branding/namespace.php
@@ -363,14 +363,20 @@ function remove_howdy_greeting( WP_Admin_Bar $wp_admin_bar ) {
  * Enqueue branding script for the post previews.
  */
 function enqueue_block_editor_branding_assets() {
+
+	global $pagenow;
+
+	$dependants = [
+		'wp-element',
+		'wp-hooks',
+	];
+
+	$dependants[] = ( $pagenow === 'widgets.php' ? 'wp-edit-widgets' : 'wp-editor' );
+
 	wp_enqueue_script(
 		'altis-branding',
 		plugin_dir_url( dirname( __FILE__, 2 ) ) . 'assets/branding.js',
-		[
-			'wp-element',
-			'wp-editor',
-			'wp-hooks',
-		],
+		$dependants,
 		false,
 		true
 	);
--- a/inc/branding/namespace.php
+++ b/inc/branding/namespace.php
@@ -363,20 +363,21 @@ function remove_howdy_greeting( WP_Admin_Bar $wp_admin_bar ) {
  * Enqueue branding script for the post previews.
  */
 function enqueue_block_editor_branding_assets() {
-
 	global $pagenow;
 
-	$dependants = [
-		'wp-element',
-		'wp-hooks',
-	];
-
-	$dependants[] = ( $pagenow === 'widgets.php' ? 'wp-edit-widgets' : 'wp-editor' );
+	// Return early when viewing the customizer or widgets screen.
+	if ( is_customize_preview() || 'widgets.php' === $pagenow ) {
+		return;
+	}
 
 	wp_enqueue_script(
 		'altis-branding',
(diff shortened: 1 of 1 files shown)


robindevitt:
@shadyvb I have created this as a draft just to share my initial implementation.

I may be overthinking this, but I feel there could be more to it rather than what I have implemented.

There may be a possibility that we'll need to factor in the new `edit-site` and `edit-navigation` packages with WP 5.9.

Initially I was thought to potentially have a function to return the script dependancy as a solution but I was hesitant to go that route, however long term it may be a better option?

The fix used in this module will be used to carry out across the rest of the modules.

inc/branding/namespace.php
```diff
@@ -363,14 +363,20 @@ function remove_howdy_greeting( WP_Admin_Bar $wp_admin_bar ) {
  * Enqueue branding script for the post previews.
  */
 function enqueue_block_editor_branding_assets() {
+
+	global $pagenow;
+
+	$dependants = [
```
shadyvb, line 369: It's dependencies rather than dependents, the other way around.

```suggestion
	$dependencies = [
```

shadyvb (commented):
Variable name update.

shadyvb:
I'd do the following to understand expected behavior:
- Check where the notice is coming from, to understand the rules.
- Check what others did to fix the same issue, by Googling the error message.

So I looked up where the notice is coming from to understand what is expected, this is where the check happens, which says `wp-editor` cannot be enqueued within either the customizer or the widget screen. ref: https://github.com/wordpress/wordpress/blob/master/wp-includes/widgets.php#L2073

Then I checked how others are handling it, Jetpack comes up: code: https://github.com/Automattic/jetpack/blob/master/projects/plugins/jetpack/modules/wpcom-block-editor/class-jetpack-wpcom-block-editor.php#L311 , ref: https://github.com/Automattic/jetpack/issues/20357



Re future-proof solution, we'll cross that bridge when we come to it. It's fine to target the current issue for now.

robindevitt:
Thanks for the feedback and direction.

I do like that way of handling it! I feel this method can be used dependant on the module it's being implemented for and the effects it may have, so I will determine it per module basis.

shadyvb:
Hmmmm, the last change is a change in direction, you were changing the dependency earlier, but now you're not loading the asset at all. Is this expected ? Do we not want to load that specific resource on customizer/widget-editor screens ?

robindevitt:
Correct, so I figured that when the script is being called but un-needed in the customizer/ widgets screen then it could return early. In cases where the asset/s are needed then changing the dependency would be preferred.

shadyvb:
@robindevitt I can approve this if you're finished with it and can confirm it is functionally tested.

robindevitt:
Thanks @shadyvb 

I'm looking for what issues come up if I return the function early, so far Im not facing any issues. Which concerns me that I am maybe missing something else.

I did identify [this issue ](https://github.com/humanmade/altis-cms/issues/532) today with Reusable Blocks so there might be a need for the dependency?

The automated tests all return successfully when running them.

shadyvb:
It doesn't seem related from looking at the actual js file we're using. you can confirm so by rolling back your change and checking the bug again.

robindevitt:
Rolling back to the previous changes Im getting the same test results.

I feel there is more value at the moment in returning early rather than changing the dependencies.

shadyvb (approved):

PR #594, opened 2022-05-30
Fix Authorship tests

roborourke:
- [x] Requires https://github.com/humanmade/altis-dev-tools/pull/370 ~so tests may need re-running once that's merged~.


roborourke:
🎉

PR #602, opened 2022-07-12
Update humanmade/asset-loader to latest

kadamwhite:
Bumps asset-loader to [latest version available on packagist.](https://packagist.org/packages/humanmade/asset-loader)


kovshenin:
Pushed authorship and altis-reusable-blocks version updates to this same branch.

PR #619, opened 2022-07-28
Bundle simple-local-avatars 2.5.0

### LINKED ISSUE #617: Simple Local Avatar Missing JS file
The version of 10Up's Simple Local Avatar we use has a JavaScript file missing. Because of changes in the plugin, it now expects to find it's JS file in a built `dist` folder. 

The git repo does not have that by default. This change seemed to arrive in version 2.2.0. The current version from the git repo, 2.5.0, has the same issue. The download from wp.org does not.

This means you cannot use the functionality form the profile page.

Possible solutions:
- Include a build step in the downloaded repo to generate the `dist` folder. Ugh! No!
- Fork the repo and add the built assets to it. 
- Require the plugin from wpackagist but still move it to the vendor folder.

kovshenin:
The 10up/simple-local-avatars package changed their release process which no longer includes built files since 2.3.0. Until the release process is fixed and the package could be used from Composer with no additional build/install steps, this PR bundles a distributable version of the plugin (2.5.0) in the lib directory.

Fixes #617


shadyvb (approved):
LGTM, but do we have a way to track upstream updates in this flow ?

roborourke:
We may need to add a `.phpcsignore` file to make linter bot happy

kovshenin:
> Do we have a way to track upstream updates in this flow ?

Not sure what our process usually is with dependencies like this, but I've subscribed to their releases and security alerts on GitHub. The issue's been added to the 2.6.0 milestone too.

PR #693, opened 2023-01-16
Explain distinction of blocking xml-rpc requests

svandragt:
Clarified the impact of disabling xml-rpc using composer.json, based on Altis Support conversations.


svandragt:
Any update for a merge? This has been causing confusion on another agency project, so there is value in this PR.

kovshenin (approved):

PR #695, opened 2023-01-19
Fix a PHP8 TypeError in 10up/simple-local-avatars

kovshenin:
There is a TypeError with PHP8 that's been fixed upstream and released in a newer version (with other changes) which we'll ship with v14. This PR backports the fix alone to v13.

See:

* https://github.com/10up/simple-local-avatars/pull/183
* https://github.com/10up/simple-local-avatars/commit/9f28dc352fae18a6fb9388c478bf7815c920b915

### DIFF
--- a/lib/simple-local-avatars/includes/class-simple-local-avatars.php
+++ b/lib/simple-local-avatars/includes/class-simple-local-avatars.php
@@ -925,7 +925,7 @@ class Simple_Local_Avatars {
 		$meta_value = array();
 
 		// set the new avatar
-		if ( is_int( $url_or_media_id + 0 ) ) {
+		if ( is_numeric( $url_or_media_id ) ) {
 			$meta_value['media_id'] = $url_or_media_id;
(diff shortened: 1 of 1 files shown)


owaincuvelier (approved):

mikelittle (commented):
I thought I saw a further fix from this (by someone at HM) as this one allows floating point numbers to pass (highly unlikely but technically possible)

kovshenin:
True, I saw your comment on the PR. Personally I wouldn't be too concerned about that as Core does:

```
$attachment_id = (int) $attachment_id;
```

Inside wp_get_attachment_url, so it shouldn't be a problem other than store a "dirty" ID in the metadata perhaps. Not sure what use case would lead to that. I don't see this as a good candidate to divert from upstream, so I'm inclined to leave as is until/unless the upstream project gets the better patch. Let me know if you strongly feel otherwise!

mikelittle (approved):
Looks good to me

PR #705, opened 2023-03-20
Reduce the style overrides for Query Monitor

johnbillion:
Some style overrides are in place for Query Monitor so that its styling more closely matches that of the purple Altis colour scheme in the admin area. However:

* The colours in these overrides are too stark (eg. dark purple for table row hover instead of pale blue)
* There are several bugs (eg. trailing icons on links aren't visible, links don't have hover states)
* The overrides aren't easily maintainable as QM evolves

These style overrides make it quite jarring to use the tool. I propose to remove all the style overrides for QM except for the Altis purple used in the main left hand side navigation menu.

## Before

Harshly contrasting table row hover state:

Barely visible trailing icon on links:

## After

Correctly styled row hover state:

Correctly styled trailing icon on links:

Altis purple remains for the main nav menu:

### DIFF
--- a/assets/admin-color-scheme.css
+++ b/assets/admin-color-scheme.css
@@ -1036,6 +1036,10 @@ textarea:focus {
 	fill: rgba( 70, 103, 222, 0.3 ) !important;
 }
 
+/**
+ * Query Monitor overrides
+ */
+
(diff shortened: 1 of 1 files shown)


kovshenin (approved):

kovshenin:
Nice one @johnbillion!

PR #776, opened 2023-12-18
Add additional signup notification email overrides

roborourke:
There were some additional notification emails that were not filtered by `wp_mail_from`.

### DIFF
--- a/inc/signup_notification/namespace.php
+++ b/inc/signup_notification/namespace.php
@@ -7,6 +7,8 @@
 
 namespace Altis\CMS\Signup_Notification;
 
+use WP_Signup_Query;
+
 /**
  * Boostrap setup to remove updates from the admin.
  */
@@ -24,10 +26,10 @@ function bootstrap() {
  * Remove the core wpmu_*_notification functions.
  */
 function remove_wpmu_blog_notifications() {
-	remove_action( 'wpmu_activate_blog', 'wpmu_welcome_notification', 10, 5 );
+	remove_action( 'wpmu_activate_blog', 'wpmu_welcome_notification', 10 );
 	remove_action( 'after_signup_site', 'wpmu_signup_blog_notification', 10 );
-	remove_action( 'wpmu_activate_user', 'wpmu_welcome_user_notification', 10, 3 );
-	remove_action( 'after_signup_user', 'wpmu_signup_user_notification', 10, 4 );
+	remove_action( 'wpmu_activate_user', 'wpmu_welcome_user_notification', 10 );
+	remove_action( 'after_signup_user', 'wpmu_signup_user_notification', 10 );
 }
 
 /**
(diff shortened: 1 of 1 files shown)


inc/signup_notification/namespace.php
```diff
@@ -160,3 +185,340 @@ function altis_signup_blog_notification( $domain, $path, $title, $user_login, $u
...
+ *
+ * @since MU (3.0.0)
+ *
+ * @param string $user_login The user's login name.
+ * @param string $user_email The user's email address.
+ * @param string $key        The activation key created in wpmu_signup_user()
+ * @param array  $meta       Optional. Signup meta data. Default empty array.
+ * @return bool
+ */
+function altis_signup_user_notification( $user_login, $user_email, $key, $meta = [] ) {
```
mikelittle, line 211: I'm guessing the standard WordPress functions had to be replicated because they don't call the filter `wp_mail_from`.

inc/signup_notification/namespace.php
```diff
@@ -160,3 +185,340 @@ function altis_signup_blog_notification( $domain, $path, $title, $user_login, $u
...
+	 * @param string $user_login User login name.
+	 * @param string $user_email User email address.
+	 * @param string $key        Activation key created in wpmu_signup_user().
+	 * @param array  $meta       Signup meta data. Default empty array.
+	 */
+	if ( ! apply_filters( 'wpmu_signup_user_notification', $user_login, $user_email, $key, $meta ) ) {
+		return false;
+	}
+
+	// Patch error in wp-user-signups passing through signup key.
```
mikelittle, line 226: It's not clear why this (and `altis_signup_blog_notification`) might be called with a blank `$key`. The WP core original doesn't seem to cater for that case. The comment mentions `wp-user-signups` but is that a function, a filter?

inc/signup_notification/namespace.php
```diff
@@ -160,3 +185,340 @@ function altis_signup_blog_notification( $domain, $path, $title, $user_login, $u
...
+		 * @param string $content    Content of the notification email.
+		 * @param string $user_login User login name.
+		 * @param string $user_email User email address.
+		 * @param string $key        Activation key created in wpmu_signup_user().
+		 * @param array  $meta       Signup meta data. Default empty array.
+		 */
+		apply_filters(
+			'wpmu_signup_user_notification_email',
+			/* translators: New user notification email. %s: Activation URL. */
+			__( "To activate your user, please click the following link:\n\n%s\n\nAfter you activate, you will receive *another email* with your login." ),
```
mikelittle, line 266: Could this be "To activate your user account..."? To be a little more friendly?

mikelittle (changes requested):
This seems like an awful lot of code for such a simple description. Is there a related issue for this?

Some comments/questions inline

inc/signup_notification/namespace.php
roborourke: Yeah if you look at the code that was already in this file that’s what it was doing, we just missed these other core notifications at the time so for a lot of stacks these emails will never successfully send if the from address doesn’t match the approved sending domains.

inc/signup_notification/namespace.php
roborourke: WP User Signups is a plugin this module provides to expose a UI for signups in the network admin. There’s a bug in that plugin which means the  key can come through empty.

inc/signup_notification/namespace.php
roborourke: Pure lift and shift from WP core. I don’t disagree though. We can make this update

mikelittle: It would be nice (and consistent with the mention of "administrator account" elsewhere). But we would lose the core translation, as we are relying on them throughout this file.

mikelittle (approved):
Looks good to me

inc/signup_notification/namespace.php
roborourke: Ah of course, good point!

roborourke:
Cheers @mikelittle!

PR #838, opened 2024-06-13
Fixes CodeCeption tests

wisyhambolu:
- Fixes https://github.com/humanmade/product-dev/issues/1525, the issue were originally failing due to rate limiting on the NGINX server on local server and was returning 429 errors. This fix adds waits at different times to avoid the rate limiting.

###TESTS ON LOCAL

```
Acceptance Tests (2) ---------------------------------------
✔ SignupsCest: Test I want to manage my network signups. (14.03s)
✔ SiteCest: Test I want to manage my network sites. (43.03s)
------------------------------------------------------------


Time: 01:25.397, Memory: 22.00 MB

OK (2 tests, 3 assertions)
Flushing the cache...
Removing test databases..

What's next?
  Try Docker Debug for seamless, persistent debugging tools in any container or image → docker debug test-root-db
  Learn more at https://docs.docker.com/go/debug-cli/
Removing headless browser container..
```

### DIFF
--- a/tests/acceptance/SignupsCest.php
+++ b/tests/acceptance/SignupsCest.php
@@ -44,6 +44,9 @@ class SignupsCest {
 		// Login with that user.
 		$I->amOnPage( '/wp-login.php?action=logout' );
 		$I->click( '//a[contains(@href,"action=logout")]' );
+
+		$I->wait(5); // wait for 5 seconds to avoid rate limiting.
+
 		$I->loginAs( 'signo', $pass );
 	}
 }
--- a/tests/acceptance/SignupsCest.php
+++ b/tests/acceptance/SignupsCest.php
@@ -45,7 +45,7 @@ class SignupsCest {
 		$I->amOnPage( '/wp-login.php?action=logout' );
 		$I->click( '//a[contains(@href,"action=logout")]' );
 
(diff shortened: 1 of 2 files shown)


mikelittle (commented):
Well done for finding this.
Is there a possibility to reduce/remove the rate limit on nginx? Otherwise, I think this means all future tests (including ones clients write) will have to incorporate these delays. 
Alternatively, do the delays need to be so long?
Tests that take too long to run tend not to get used.

kovshenin (approved):

kovshenin:
Change looks fine. I agree re. long tests though, I think it's worth removing rate limiting in Local Server, it is not useful at all.

joehoyle:
We can override the rate limits in the nginx container per https://github.com/humanmade/docker-wordpress-nginx/blob/master/docker-entrypoint.d/config-replacements.sh, we'd need to just define the env vars in https://github.com/humanmade/altis-local-server/blob/master/inc/composer/class-docker-compose-generator.php#L337

wisyhambolu:
Thanks for the feedback everyone, I'll create a PR to override the rate limits in the nginx container.

wisyhambolu:
Closing in favour of https://github.com/humanmade/altis-local-server/pull/698 and https://github.com/humanmade/docker-wordpress-nginx/pull/96

PR #855, opened 2024-07-29
Filter causes PHP Warning. Remove it.

mikelittle:
The "Howdy" greeting in the admin bar this filter is meant to replace is now set via JavaScript in the admin bar, so this no longer works. Instead, it generates a PHP warning because the title is null.

Addresses first par t of https://github.com/humanmade/product-dev/issues/1608
Relates to https://github.com/humanmade/product-dev/issues/1602


mikelittle:
Turns out it was priority changes that stopped the original action working.
Closing in favour of https://github.com/humanmade/altis-cms/pull/856

PR #856, opened 2024-07-31
Fix Howdy removal for WordPress 6.6

kovshenin:
Core 6.6 moved some priorities around for admin bar nodes. Our old priority can't find the my-account node because it's added at a later priority.

See: https://core.trac.wordpress.org/changeset/58759


mikelittle (approved):
Works. 
Thanks for spotting this. I closed my other PR.

PR #890, opened 2025-03-05
Adds a local server specific favicon to Altis

mikelittle:
This adds a local server specific favicon to Altis. This is useful for local development environments to easily distinguish between local and remote servers.

Fixes: https://github.com/humanmade/altis-local-server/issues/683

I have used the Altis Logo with a blue background as the Local-Server-specific version.
CleanShot 2025-03-05 at 18 07 03


rmccue:
Checking with Pawel on the design for this one - we might have a better variant we can use here.

rmccue (approved):

PR #892, opened 2025-04-09
Update the minimum PHP version to 8.2

mikelittle:
Update the minimum PHP version to that which Altis supports. This will allow dependabot to run successfully.

Addresses: humanmade/product-dev#1735


wisyhambolu (approved):
Looks good

PR #893, opened 2025-04-30
Upgrade to WordPress 6.8

mikelittle:
Upgrades to use WordPress 6.8
Also removed the `roots/wp-password-bcrypt` library as it is no longer needed.

Addresses: https://github.com/humanmade/product-dev/issues/1768


wisyhambolu (approved):
Looks good

PR #903, opened 2025-08-29
Update simple-local-avatars version to 2.8.5

roborourke:
Update the plugin dependency to use the wpackagist version.

This resolves the issue that the plugin expects the `10up/wp-compat-validation-tool` to be local to the plugin code and modified to have a different namespace.

The CMS module needs to initialise the plugin from the `WP_PLUGIN_DIR` directory.

- Resolves https://github.com/humanmade/product-dev/issues/1834

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
     "humanmade/clean-html": "^2.0.0",
     "humanmade/altis-reusable-blocks": "~0.2.4",
     "humanmade/asset-loader": "^0.6.4",
-    "wpackagist-plugin/simple-local-avatars": "~2.8.5"
+    "10up/simple-local-avatars": "~2.8.5"
   },
   "autoload": {
     "files": [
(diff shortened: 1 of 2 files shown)


maugelves (approved):

mikelittle:
The multiple change logs for this don't look too bad. But for such a large leap, we need to do some testing.
I'll follow up on that.

roborourke:
Yeah sorry, you should remove the back port labels and I’ll leave it to your discretion

mikelittle (changes requested):
I can't get this to install in product dev.
Since 2.8.0 it now requires `10up/wp-compat-validation-tool` at `dev-trunk` which is not published on Packagist.

If I manually add the git repo as a source (not gonna happen for production), it still won't install as that needs an incompatible version of "composer/installers"

Have you got this to work on a project?

goldenapples:
:wave: Is this ready to release and backport? I'm responding to a client's security scan and need to know whether to replace the package or expect an update.

PR #918, opened 2026-01-29
Fix broken test.

mikelittle:
The search for "Add Site" to click finds the menu item first. But is is hidden because the menu is not expanded on the small screen size the tests run under. Changed the test to use the unique ID the button has.

CleanShot 2026-01-29 at 16 52 42

During the test, the menu is not visible
SiteCest testSitesManagementActions fail
So the code can't "click" it.

- Fixes https://github.com/humanmade/product-dev/issues/1815


vladislavhmn (approved):
```
Acceptance Tests (2) ---------------------------------------
✔ SignupsCest: Test I want to manage my network signups. (5.25s)
✔ SiteCest: Test I want to manage my network sites. (19.42s)
------------------------------------------------------------
Time: 00:40.050, Memory: 22.00 MB

OK (2 tests, 4 assertions)
```

PR #940, opened 2026-03-25
Remove altis-reusable-blocks module in favour of core synced patterns

mikelittle:
WordPress 6.3+ provides native synced patterns that replace the functionality previously provided by humanmade/altis-reusable-blocks. Remove the composer dependency, plugin loading, and config gate. Update docs to reference core synced patterns instead.

See https://github.com/humanmade/product-dev/issues/2046


mikelittle:
Created as a draft, as this needs to be announced as a deprecation first.

wisyhambolu (approved):
Looks good!

PR #942, opened 2026-04-16
Revert "[Backport v25-branch] Remove plaintext passwords from multisite activation flow"

mikelittle:
Reverts humanmade/altis-cms#926

This is a behaviour change and should not have gone to this version


wisyhambolu (approved):
Looks good

PR #952, opened 2026-05-21
Update GHA reusable workflow ref

mikelittle:
Pin to altis-dev-tools tip past #1056 (artifact name + module-config input + Packagist bypass + other follow-ups).


ferschubert-hm (approved):
LGTM

